PR #17, opened 2022-05-23
uiboard integration

ene9ba:


.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "ROS/src/open_mower_ros"]
 	path = ROS/src/open_mower_ros
-	url = https://github.com/ClemensElflein/open_mower_ros.git
+	url = https://github.com/ene9ba/open_mower_ros.git
```
conradkirschner, line 3: I think it should show to the main repo and not a fork.

ClemensElflein: True, thank you for checking! I'll need to do some minor adjustments as well to this PR before merging, I promised @ene9ba that I'll be doing the changes before merging 🙂

PR #21, opened 2022-06-23
Automatic generation of bom/render/gerbers

rfvermut:
I updated the scripts, now it runs CI on every commit that does DRC/ERC
failing due to ERC https://github.com/rfvermut/OpenMower/runs/7030384610?check_suite_focus=true 

Release now includes a render of a board with git tag as version on a silkscreen: https://github.com/rfvermut/OpenMower/releases/tag/v0.9.10

also I added the script that regenerates GH pages with a navigator of all artifacts (except ibom, somehow it dislikes it)
https://rfvermut.github.io/OpenMower


ClemensElflein:
Nice, thank you! I'll merge that and then I have to fix the DRC and ERC issues, but I think it's a good idea to have it. I will copy that workflow to the xESCs as well

PR #22, opened 2022-07-05
feat(GH): pick and place files

rfvermut:
add Position files to release bundle


ClemensElflein:
Thank you for the PR, having the pick and place files as well is a good idea.

PR #24, opened 2022-07-20
temporary fix - fuse holder PN

vamatis:
fixed the PN for the fuse holder (temp fix)


rfvermut:
1. It has `.DS_Store` file which probably is completely wrong to be there
2. It also renames bunch of other references - R19, Q16. This is probably legit, but then it's not what PR/commit name says.

rfvermut:
Actually I'm getting the same changes (R19, Q16) automatically generated when I'm doing completely unrelated changes. Probably it's some KiCAD magic beyond my comprehension.

vamatis:
Hi,

thanks for the comments and sorry for the late reply.
regarding the .DS_Store, it's probably because I use a mac and these dumb machines tend to generate crazy files like that for no good reason... I'll add them to the .gitignore.

As for the references, "glad" you had the same issue haha because I had no idea where that came from, as I only made the modification listed...

Weird stuff.. But I'll have to check because changing the resistors name at this point would probably cause a great mess

rfvermut:
I think it's related to the fact that `LEVEL_SHIFTER` subsheet is reused several times with multiple Q/R value variations.

rfvermut:
I tried reading things (https://www.re-innovation.co.uk/docs/multiple-sheet-schematics-in-kicad/), it only suggests to re-run annotations. Which might by a good thing?.. Or not?.. @ClemensElflein  ?

ClemensElflein:
Not sure what that means, I usually just press F8 to apply schematic changes to the PCB. I never run annotation manually. Parts between schematic and PCB are matched using a unique identifier as far as I know.

rfvermut:
Ok, let me give a reproducible example.

Start with clean slate
```
# git reset --hard upstream/main
HEAD is now at c7388a3 Revert "feat(GH): combine and reuse workflows (#26)"
 
# git status
On branch main
Your branch is up to date with 'upstream/main'.
```

Now launch KiCad, open Schematic Editor and **without making any changes** just File -> Save it. You get the same weird changes that popped up in current PR
```
# git status    
On branch main
Your branch is up to date with 'upstream/main'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   Hardware/OpenMowerMainboard/LEVEL_SHIFTER.kicad_sch
        modified:   Hardware/OpenMowerMainboard/opamps.kicad_sch

# git diff | cat
diff --git a/Hardware/OpenMowerMainboard/LEVEL_SHIFTER.kicad_sch b/Hardware/OpenMowerMainboard/LEVEL_SHIFTER.kicad_sch
index 180176b..c62e675 100644
--- a/Hardware/OpenMowerMainboard/LEVEL_SHIFTER.kicad_sch
+++ b/Hardware/OpenMowerMainboard/LEVEL_SHIFTER.kicad_sch
@@ -270,10 +270,10 @@
     (uuid c10ace36-a93c-4c08-ac75-059ef9e1f71c)
   )

-  (symbol (lib_id "Transistor_FET:BSS138") (at 114.3 76.2 270)
+  (symbol (lib_id "Transistor_FET:BSS138") (at 114.3 76.2 270) (unit 1)
     (in_bom yes) (on_board yes)
     (uuid 00000000-0000-0000-0000-0000613371aa)
-    (property "Reference" "Q16" (id 0) (at 114.3 82.5246 90))
+    (property "Reference" "Q13" (id 0) (at 114.3 82.5246 90))
 <more lines...>
```

My best guess is that schematics were changed manually? directly in sch files? and got out of sync. So we just need to commit those changes and become in sync.

----

Also, open PCB editor and press F8. It will add `text_property` which probably is my mistake and that one should have been part of CI commit.
```
# git status
On branch main
Your branch is up to date with 'upstream/main'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   Hardware/OpenMowerMainboard/LEVEL_SHIFTER.kicad_sch
        modified:   Hardware/OpenMowerMainboard/OpenMowerMainboard.kicad_pcb
        modified:   Hardware/OpenMowerMainboard/opamps.kicad_sch
```

ClemensElflein:
@vamatis I'd add the ali link in #30, so that we don't get merge issues with the other changes in the files.

vamatis:
Sure no prob! I'll just close this one then !

PR #28, opened 2022-08-09
Fixes from Vermut's 1st batch

rfvermut:
- fix(soundmodule): female socket instead of male headers
- fix(motorsensor): J10 was upside-down
- feat(pcb): rotate J10 and move lower, soundboard pin sockets


ghost:
Hi, 
J10 has to go the opposite direction of the adjacent J4?
J10_Fix

I took the original one out and want to make sure before I replace.

and the male audio module connector should be swapped for a female pin header?
Fix_Audio_MtoF

rfvermut:
> J10
This is how it supposed to look like
image

rfvermut:
These two need to become females. Or at least start identify themselves as.
image

ghost:
Messed up a small bit with my fix. Should have just popped the plastic housing off the pins and flipped it around 180.

Only other thing I noticed was that 1 ground pin of the xESC2040 connectors at the edge needed some solder. 
 
Besides that, boards look really nice. Thanks

rfvermut:
> Messed up a small bit with my fix. Should have just popped the plastic housing off the pins and flipped it around 180.

Can you tell me more? Do you mean there is a solderless way to fix it?

ghost:
Will have to check, I've orders some 2.0mm JST connectors and sockets from Ali. I only had the 2.54mm ones.
https://www.aliexpress.com/item/32908472053.html

(What's the name of that J10 connector type with the locking clip in the picture you posted showing the fix? May have ordered wrong sockets.)


245vs2wclip

My thinking was that the plastic could have been careful pried off the pins and flipped around to be in the right orientation.  (Not sure this will work now since you said J10 has to be moved lower, maybe there is insufficient space. )
180socket It would be a very close fit.


If there's no space I can perhaps wire it like Clemen's board, with some wire to a socket.

headerfix

ClemensElflein (approved):
Looks good to me.
Changes I noticed:
- Rotated mower connector-
- Some silkscreen changes (female pin headers, corner of neopixel)

PR #30, opened 2022-08-14
More LCSC codes

rfvermut:


Hardware/OpenMowerMainboard/RobotConnectors.kicad_sch
```diff
@@ -3287,6 +3290,12 @@
     (property "Stock_PN" "FH-FC-202-24" (id 6) (at 118.745 68.58 0)
       (effects (font (size 1.27 1.27)) hide)
     )
+    (property "Description" "LCSC links to resettable PTC" (id 7) (at 118.745 68.58 0)
+      (effects (font (size 1.27 1.27)) hide)
+    )
+    (property "LCSC" "C369104" (id 8) (at 118.745 68.58 0)
```
ClemensElflein: Not sure if that is the right fuse to use. The fuse trips at 10A for 11 seconds. That's about 300W of power for 11 seconds, something is going to get hot.

ClemensElflein (commented):
Looks good to me except for the fuse. I'm personally not a fan of these slow, resetable fuses. I can't see the benefit in this case, if any component gets 300W for 11sec you'll need to open the mower to repair anyways; Opening to "just" replace the fuse makes more sense to me.

Hardware/OpenMowerMainboard/RobotConnectors.kicad_sch
rfvermut: I'll find something faster in catalogue.

rfvermut:
Reverted fuses back to Ali fuse holders.

Merged #24 changes in

rfvermut:
also added 3D model path fixes to this PR, now more components are rendered
image

PR #32, opened 2022-08-23
Multi IMU

rfvermut:
See commits for changes


ClemensElflein:
Thank you for the PR!

- There are some issues I found with the connections of the Pico: GP6 is SCK, but used as MOSI, 7 is MOSI and it's used as clock and so on.
- Could we make the naming for GPIO8 and GPIO9 in a way that we can see that these are used in duplicate functions (e.g. SP_MISO_UART_TX for GPIO8)
- Could you please rename GPIO16 and GPIO17 with something like SOUND_TX and SOUND_RX, it's a serial connection now. It used to be for another sound module which used another protocol.

Everything else looks good to me.

ClemensElflein:
+ If you're already rerouting the Pico, could you update the footprint for the Pico to this one, so that we can use the Pico version with Wifi (request in the community): https://forum.kicad.info/t/are-there-pi-pico-library-files-available-for-kicad-6/35844/12

ClemensElflein:
Thank you for the update, looks good to me!

PR #35, opened 2022-09-11
feat(imu): ttl/i2c selector for WT901

rfvermut:
+ moved voltage selector for sound

image

image


ClemensElflein:
Looks good, thanks for the PR

PR #45, opened 2022-11-25
rpi boot button, wt901 rearrangement and fixes

rfvermut:
image

image
image


rfvermut:
PR is totally wrong because I forgot about PCB holder hooks that going to overlap places I've used

rfvermut:
updated PCB (see render in description)

PR #61, opened 2023-04-11
Feature/cover UI yfc500 1.3.0

Apehaenger:
Hi Clemens,

here's now how I did it with your code to support the original YardForce Classic 500 CoverUI (STM32F030R8T6 variant).
At the moment only the LED stuff is integrated, but I wanted to give you a look to see if it's worth to go further this direction.

My main goal was to integrate it without polluting or changing your code to much!

I mainly did it in the following structure/way:
1. "CoverUI/yfc500" has all relevant CoverUI for STM32 code
2. "CoverUI/yfc500/stm32cube" contains the auto-generated STM32CubeIDE hardware definition/header without further modification
3. "CoverUI/yfc500/main.hpp" is the main and only header file which need to be included in your main.cpp
4. "CoverUI/yfc500/LEDcontrol.h|cpp" is the replacement for your LEDcontrol and statemachine

But for sure I had to pollute your main.cpp with a couple of ```#ifdef HW_YFC500 ...``` 

I mainly used my picoprobe during development (less wires on table because his additional UART). As I'm pretty sure that you also have a picoprobe, I didn't made an ST-Link env entry in platformio.ini. But if you prefer ST-Link I can add it.

It also compiles still well with Pico-SDK but because I don't have an OM-CoverUI, I couldn't test it.


Apehaenger:
Hi Clemens!

Well I'm **nearly** done now. Generic button support is also in now.

But I've a remaining understanding/usage question:
If I understand it right, `Force_LED_on()` and `Force_LED_off()` overrules the current/saved LED state of `uint64_t LED_activity`.
In the current state, yet only used by `Blink_LED()`, which in turn is used as long-press- button detection/signalling (for buttons with a LED).

What I'm in doubt is, why `Blink_LED()` does the two 10*10ms loops:
```
void Blink_LED(PIO pioBlock, int statemachine, int led)
{

    printf("flashing led %d\n", led);
    Force_LED_off(led, false);
    Force_LED_on(led, true);
    for (int i = 0; i < 10; i++)
    {
        LEDs_refresh(pioBlock, statemachine);
        sleep_ms(10);
    }
    Force_LED_off(led, true);
    Force_LED_on(led, false);
    for (int i = 0; i < 10; i++)
    {
        LEDs_refresh(pioBlock, statemachine);
        sleep_ms(10);
    }
    // stop with forced off
    Force_LED_off(led, true);
    Force_LED_on(led, false);
}
```

What I guess is, you did it for a remaining blink rate of the possible other LED's in blink state?

So, if I understand the logic right, long-pressing a button (with a LED), simply overrules his current state and let him blink fast (100ms) as long it's pressed?

ClemensElflein:
Yes you got it right I think. Basically the code forces one LED on but continues to refresh the state machine (so that the blinking for other LEDs still continues to work and doesn't look frozen / glitchy) for 100ms. Then the LED is forced off and the state machine is updated for 100ms.

The idea is to give the user feedback on what a long press is.

Pressing longer than the long press timeout will blink the LED once and continuing to hold for another long press duration will flash the LED twice.

ClemensElflein:
Moved CoverUI to a new repo: https://github.com/ClemensElflein/CoverUI

Closing this

PR #64, opened 2023-04-23
Create mowareareader.py

0815eddi:
mowareareader: added obstacles


ClemensElflein:
Could you please add a small readme.md file to this, so that people know how to use it?

0815eddi:
In the meantime there are other map editing tools like rosbagReader.py. So maybe it 's better first to compare the tools and than focus on one of them for further development. In the next days I will ask on discord to get an overwiew.

PR #65, opened 2023-05-11
Stock-CoverUI integration proposal

Apehaenger:
Hi Clemens!

This is only a proposal how I would/could integrated the Stock-UI packets into the LowLevel code.
~~**Not compiled nor tested!** Simply written down to show how I would do it.~~

In this proposal I've chosen the way via two additional Stock-UI specific global vars, as by this I had the impression to get easiest integration into the existing code without disturbing logic or understanding.

~~What I'm unsure yet is:~~
~~- What get detected if no rain-sensor is connected to OM's mainboard?
  Same as with the halls? (not connected = no emergency/rain flag)~~

### DIFF
--- a/Firmware/LowLevel/src/main.cpp
+++ b/Firmware/LowLevel/src/main.cpp
@@ -88,6 +88,10 @@ unsigned long last_UILED_millis = 0;
 unsigned long lift_emergency_started = 0;
 unsigned long button_emergency_started = 0;
 
+// Stock UI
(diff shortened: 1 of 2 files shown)


Apehaenger:
Compiles well and work as expected ;-)

ClemensElflein:
Thanks for the PR, looks good to me.

Apehaenger:
That's cool, quite thanks!

PR #71, opened 2023-06-26
Modify wheel lift logic for tilt & lift scenarios. 

kevinmce:
Addition of single wheel lift as tilt scenario with increased time before emergency. 
Previous lift emergency still triggers in 500 milliseconds like previously but the single wheel lift has an extended trigger time of 2500 milliseconds. 

This keeps the safety of a real lift scenario but also allows for rough terrain in your garden.

### DIFF
--- a/Firmware/LowLevel/src/main.cpp
+++ b/Firmware/LowLevel/src/main.cpp
@@ -31,7 +31,8 @@
 #define STATUS_CYCLETIME 100      // cycletime for refresh analog and digital Statusvalues
 #define UI_SET_LED_CYCLETIME 1000 // cycletime for refresh UI status LEDs
 
-#define LIFT_EMERGENCY_MILLIS 500  // Time for wheels to be lifted in order to count as emergency. This is to filter uneven ground.
+#define TILT_EMERGENCY_MILLIS 500  // Time for wheels to be lifted in order to count as emergency. This is to filter uneven ground.
+#define LIFT_EMERGENCY_MILLIS 2000  // Time for one wheel to be lifted in order to count as emergency. This is to filter uneven ground.
 #define BUTTON_EMERGENCY_MILLIS 20 // Time for button emergency to activate. This is to debounce the button if triggered on bumpy surfaces
 
 // Define to stream debugging messages via USB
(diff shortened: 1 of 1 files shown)


Firmware/LowLevel/src/main.cpp
```diff
@@ -31,7 +31,8 @@
 #define STATUS_CYCLETIME 100      // cycletime for refresh analog and digital Statusvalues
 #define UI_SET_LED_CYCLETIME 1000 // cycletime for refresh UI status LEDs
 
-#define LIFT_EMERGENCY_MILLIS 500  // Time for wheels to be lifted in order to count as emergency. This is to filter uneven ground.
+#define TILT_EMERGENCY_MILLIS 2500  // Time for wheels to be lifted in order to count as emergency. This is to filter uneven ground.
+#define LIFT_EMERGENCY_MILLIS 500  // Time for one wheel to be lifted in order to count as emergency. This is to filter uneven ground.
```
ClemensElflein: I think the comments are the wrong way around. Also with the new change I think we can lower the lift limit even more that can stop instantly when the mower drives over a cliff 

```suggestion
#define TILT_EMERGENCY_MILLIS 2500  // Time for a single wheel to be lifted in order to count as emergency. This is to filter uneven ground.
#define LIFT_EMERGENCY_MILLIS 100  // Time for both wheels to be lifted in order to count as emergency. This is to filter uneven ground.
```

Firmware/LowLevel/src/main.cpp
```diff
@@ -509,7 +530,7 @@ void onPacketReceived(const uint8_t *buffer, size_t size) {
 
 // returns true, if it's a good idea to charge the battery (current, voltages, ...)
 bool checkShouldCharge() {
-    return status_message.v_charge < 30.0 && status_message.charging_current < 1.5 && status_message.v_battery < 29.0;
+    return status_message.v_charge < 30.0 && status_message.charging_current < 2 && status_message.v_battery < 29.0;
```
ClemensElflein: please revert this change
```suggestion
    return status_message.v_charge < 30.0 && status_message.charging_current < 1.5 && status_message.v_battery < 29.0;
```

PR #72, opened 2023-07-03
Fix emergency bit confusion

Apehaenger:
Recognized that @kevinmce also had some emergency-bit confusion in his cool #71, like me (during my Stock-CoverUI development).

Tried to clear that up by evaluating the whole chain from Hall-x labels on the PCB, PIN_EMERGENCY_x definitions, up to the emergency logic, with comparing the the actual [doc](https://openmower.de/docs/robot-assembly/assemble-the-robot/#step-4-connect-cables) on a 0.9.x OM-MB, and also cross checked PCB Hall- labeling with PIN_EMERGENCY_x definition on the 0.13 schematics.
Beside my further tries to make that part some easier to read, I did not changed any emergency bit position, but corrected the (in my point of view) wrong comments.

Because testing take mostly longer than coding, I would like to ask before, what do you think about my adjustements and if you're interested to merge, if it got tested?


ClemensElflein:
Good idea, AFAICT, this is already part of #80? Should we close this PR then?

Apehaenger:
No. Only the #defines are in #80 the logic change and the corrected emergency comments are here.

Firmware/LowLevel/src/datatypes.h
```diff
@@ -52,10 +65,10 @@ struct ll_status {
     float uss_ranges_m[5];
     // Emergency bitmask:
     // Bit 0: Emergency latch
-    // Bit 1: Emergency 0 active
-    // Bit 2: Emergency 1 active
-    // Bit 3: Emergency 2 active
-    // Bit 4: Emergency 3 active
+    // Bit 1: Emergency/Hall 3 active
```
Apehaenger: This comments seem to be the culprit of all emergency confusions

Apehaenger:
Corrected online. Not validated yet. To late now :-(
Will do tomorrow and add a further comment.

Apehaenger:
1. Checked once more 0.9.x and 0.13.x schematics as well as PCB drawings
2. Switched to bitwise emergency processing as it save some if's, assignments as well as simplifies Stock-CoverUI integration
3. Unfortunately changed processing order of stop, tilt and lift handling with the result that the diff view look somehow confusing
4. Tested all Hall1-4 inputs and cases (stop, tilt, lift) with a 0.9.x on my table as well as with a Stock-CoverUI. For the latter I couldn't test the Hall Inputs (no cables), but was able to test the Stop buttons

ClemensElflein (approved):

PR #74, opened 2023-08-07
Initialize PIN_ENABLE_CHARGE pin to HIGH to reset BMS undervoltage

olliewalsh:
Provides a simple method to reset BMS undervoltage protection:
Switch off main power switch with mower out of dock.
Place mower in dock.
Switch on main power switch.

Should be pretty safe to do this, checkShouldCharge will be run immediately and disable charging if any of the values are off.


ClemensElflein:
Good point! Thanks for the PR!

PR #75, opened 2023-08-07
Improve ADC accuracy

olliewalsh:
Use the charge current ADC when undocked to determine ADC offset, tweak the resistors values, and disable SMPS powersaving during ADC.

With this I get v_battery pretty close to the measured value

### DIFF
--- a/Firmware/LowLevel/src/main.cpp
+++ b/Firmware/LowLevel/src/main.cpp
@@ -694,7 +694,11 @@ void loop() {
 
 
         // If mowing use charge current ADC to determine adc offset
-        if( last_high_level_state.current_mode == HighLevelMode::MODE_AUTONOMOUS && last_high_level_state.gps_quality != 0 ) {
+        if(
+                ROS_running &&
+                last_high_level_state.current_mode == HighLevelMode::MODE_AUTONOMOUS &&
+                last_high_level_state.gps_quality != 0
+            ) {
             adc_offset_samples[next_adc_offset_sample++] = (float)analogRead(PIN_ANALOG_CHARGE_VOLTAGE);
             next_adc_offset_sample %= 20;
 
--- a/Firmware/LowLevel/src/main.cpp
+++ b/Firmware/LowLevel/src/main.cpp
@@ -62,6 +62,10 @@ SerialPIO uiSerial(PIN_UI_TX, PIN_UI_RX, 250);
 #define R_SHUNT 0.003f
 #define CURRENT_SENSE_GAIN 100.0f
 
+int next_adc_offset_sample = 0;
+float adc_offset_samples[20] = {0};
+float adc_offset = 0.0f;
+
 #define BATT_ABS_MAX 28.7f
 #define BATT_ABS_Min 21.7f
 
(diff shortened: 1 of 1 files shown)


olliewalsh:
Note: I'm using a picow, interested in feedback from someone using a pico as the 3.3v supply is slightly different

Firmware/LowLevel/src/main.cpp
```diff
@@ -56,11 +56,15 @@ SerialPIO uiSerial(PIN_UI_TX, PIN_UI_RX, 250);
 /**
  * @brief Some hardware parameters
  */
-#define VIN_R1 10000.0f
-#define VIN_R2 1000.0f
+#define VIN_R1 9980.0f
+#define VIN_R2 998.0f
```
ClemensElflein: Where do these come from? I assume these will be slightly different for every board (10k 1%, 1k 1%), so we should have them at 10k and 1k

ClemensElflein:
Thank you for the PR! Looks good!

- Should we use the "raw" voltage for battery overvoltage protection instead of the corrected one? I'm thinking about what happens if the offset is "wrong" (e.g. is at 3V) and breaks overvoltage shutdown. For the overvoltage shutdown the noise and offset don't really matter I think

- Do you know what the average error is you get? I assume it's <1V. Should we lower the 3V check?

Firmware/LowLevel/src/main.cpp
```diff
@@ -601,16 +605,32 @@ void loop() {
     if (now - last_status_update_millis > STATUS_CYCLETIME) {
         updateNeopixel();
 
+        // Disable power saving during ADC
+        digitalWrite(PIN_SMPS_POWERSAVE, HIGH);
```
olliewalsh: In spite of what the datasheets says this is actually worse on my mower and seems to be a general issue with all picos from https://pico-adc.markomo.me/

olliewalsh:
> Thank you for the PR! Looks good!
> 
> * Should we use the "raw" voltage for battery overvoltage protection instead of the corrected one? I'm thinking about what happens if the offset is "wrong" (e.g. is at 3V) and breaks overvoltage shutdown. For the overvoltage shutdown the noise and offset don't really matter I think
> * Do you know what the average error is you get? I assume it's <1V. Should we lower the 3V check?

Replaced the logic to check the high-level mode && GPS to determine that it's mowing, ensures it doesn't try to adjust the offset while docking.
Switched to v_charge instead of charge_current adc as current is affected by regen AFAICT while v_charge is not (behind a diode).
Limited offset to +-3% to ensure it's sane

pelle-c:
I haven't tested this PR in my mower, but I did some measurements to see if it could work since I thought it was a great idea to have self calibration.

My 0.13.2 mb is somewhat modified since I had issues getting stable readings - I now have a capacitor over 3.3V and an "external" voltage reference to ADC_VREF.
When my mower is not connected to the charger, the reported charge voltage is ~0.45V. If I short the related pin to GND, I get ~0.13V. 
If I understand this PR correctly, the charge voltage is measured and "averaged" while mowing. This value is then used to offset the measured voltages. In my case, that would mean that the voltage was offset with -0.45 instead of -0.13, which would lead to readings that are ~0.3V off from the real value. 

From reading PRs here, and postings on discord, I've got the impression that you have some variations in the ADC errors. Like for me on this one, does the same apply for all boards or is just on 0.13.2, or only mine? 
Guess it's hard to tell without gathering info from users.

PR #76, opened 2023-08-07
Allow TILT/LIFT emergency to be disabled completely

olliewalsh:


ClemensElflein:
let's set this up together with this:
https://github.com/ClemensElflein/open_mower_ros/issues/50#issuecomment-1669281706

PR #78, opened 2023-10-12
Better support for CoverUI's "with Display" Proposal

Apehaenger:
# Issue

When porting a stock CoverUI with some kind of display (instead of simple LEDs), it's currently required to translate the LED states (on, off, blink_slow, blink_fast) back to some kind a displayable value.
This is somehow cumbersome as well as inaccurate (i.e. charging state is currently represented by one LED).

Because updating the CoverUI is painful (you've to open the mower), it's required to find a solution which doesn't influence current implementations in any way.

# Approach

Implement some kind of request which has to be initiated by the CoverUI's FW which is interested in other data than the current Set_LEDs packet.

# Implementation

For this, I added the msg_event_subscribe packet, which can be requested by the CoverUI to "subscribe" to specific data. Default to Set_LEDs packet, so that current installations work as before.

# Summary

This is a proposal. Nothing tested yet.
Like to know you thoughts.
As I expect to get two more stock CoverUI's the next days, and one will have a display again, I like to find a solution which is clear and stable for some longer time.

### DIFF
--- a/Firmware/LowLevel/src/main.cpp
+++ b/Firmware/LowLevel/src/main.cpp
@@ -29,7 +29,6 @@
 
 #define IMU_CYCLETIME 20              // cycletime for refresh IMU data
 #define STATUS_CYCLETIME 100          // cycletime for refresh analog and digital Statusvalues
-#define UI_SET_LED_CYCLETIME 1000     // cycletime for refresh UI status LEDs
 #define UI_GET_VERSION_CYCLETIME 5000 // cycletime for UI Get_Version request (UI available check)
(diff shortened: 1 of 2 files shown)


Apehaenger:
Implemented CoverUI- subscription functionality in Stock- SAxPRO model, and tested with and without subscription, also with C500B and RM-ECOW-V1.0.0 (NX80i) CoverUIs.
Works as expected.
Would be cool if it could be merged as it's required now for SAxPRO and will also be required by (future) NX100i CoverUI port.

ClemensElflein:
LGTM, thank you for the PR.

PR #79, opened 2023-10-30
Extra caps and silkscreen

rfvermut:
Add missing big caps for ESCs
Upgrade to Kicad7
Put back sound resistors


ClemensElflein:
Looks good to me, I have converted to KiCAD8, but there is no KiBOT for V8 yet, so we can merge later: https://github.com/ClemensElflein/OpenMower/tree/rfvermut-extra-caps-and-silkscreen

PR #88, opened 2024-04-05
Implement flexible config req- & response- packages. See PR79 of open…

Apehaenger:
Here's how I would implement my newest changes of https://github.com/ClemensElflein/open_mower_ros/pull/79

The main change is now, that LL would send and request a config packet whenever ROS_running is (back) true.
By this we would cover LL as well as HL restarts.


ClemensElflein:
LGTM

PR #90, opened 2024-04-19
Add 0_9_X_WT901 binary

Apehaenger:
I missed the binary for my HW combination ;-)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -82,6 +82,10 @@ jobs:
           cp Firmware/LowLevel/.pio/build/0_9_X_MPU9250/firmware.elf ./artifacts/0_9_X_MPU9250
           cp Firmware/LowLevel/.pio/build/0_9_X_MPU9250/firmware.uf2 ./artifacts/0_9_X_MPU9250
 
+          mkdir ./artifacts/0_9_X_WT901
+          cp Firmware/LowLevel/.pio/build/0_9_X_WT901/firmware.elf ./artifacts/0_9_X_WT901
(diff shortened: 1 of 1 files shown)


Apehaenger:
In addition:
We don't have a `0_12_X_LSM6DSO` build target.
Instead of `0_12_X` get build with `+<imu/LSM6DSO/>`.

But I don't know the historic reason why 0_12_X get copied as 0_12_X_LSM6DSO (that's why I only commenting it)

https://github.com/ClemensElflein/OpenMower/blob/14f4bf2bab4ec8d0ff1c63968216e55b30bfc150/.github/workflows/ci.yaml#L57-L63

vermut:
> In addition: We don't have a `0_12_X_LSM6DSO` build target. Instead of `0_12_X` get build with `+<imu/LSM6DSO/>`.
> 
> But I don't know the historic reason why 0_12_X get copied as 0_12_X_LSM6DSO (that's why I only commenting it)

I think we were moving from multiple IMU towards standardized boards/kits, so 0_12_X automatically assumed LSM6DSO. The same for 0_13_X.

```
# Supported values as of today:
# 0_13_X: Use this if you have 0.13.x mainboard with LSM6DSOTR (default).
# 0_12_X_LSM6DSO: Use this if you have an LSM6DSOTR and have a 0.12.x mainboard.
# 0_11_X_WT901: Use this if you have an WT901 and have a 0.11.x mainboard.
# 0_10_X_WT901: Use this if you have an WT901 and have a 0.10.x mainboard.
```

vermut:
Regarding your PR - you saying that we are building the binary but not packaging that?

Apehaenger:
Exactly. It's in PatformIO env and compiles well, but didn't get packed.

PR #92, opened 2024-05-03
Fix motor configs

olliewalsh:
Batter/motor current limits were the wrong way around. Also disable back-off at lower voltage.


ClemensElflein:
Thank you for the PR, I have read the discussion on Discord. Weird that noone noticed that before..?

PR #96, opened 2024-05-17
Bugfix/wt901 via SerialPIO

Apehaenger:
Got WT901 via SerialPIO now working reliable. Also tested if WT901_INSTEAD_OF_SOUND still works = yes.
It's still draft as I wait for Deyan to test it also (he's on the road for a week), but like to comment before forget ;-)

Neopixel:
For easier communication-error identification (as well as helper during my debugging), I added a magenta color to Neopixel which flashes whenever an overflow or packet error occurs (Pico's LED was a bad idea by me, as it's hidden by RTK2B).
If accepted and merged, we probably should also add other SerialPIO devices to check if they also get affected by the SerialPIO timing issue. "Sound" probably also, but the sound packets are much smaller. CoverUI?!
For this I had to slow down the normal Emergency flash rate by one more bit. Without this, visual identification wasn't possible and you quickly got blind :-/

This PR also contain a tiny nv_config fix for a compiler warning, as long as #80 isn't merged.

### DIFF
--- a/Firmware/LowLevel/lib/JY901_SERIAL/JY901.cpp
+++ b/Firmware/LowLevel/lib/JY901_SERIAL/JY901.cpp
@@ -13,10 +13,10 @@ CJY901 ::CJY901(SerialPIO *serial)
 void CJY901::begin(unsigned long baudrate) {
 	serial->begin(baudrate);
 	delay(1000);
-	/*uint8_t unlock[] = {0xFF,0xF0,0xF0,0xF0,0xF0}; // undocumented / unknown?
+	uint8_t unlock[] = {0xFF,0xF0,0xF0,0xF0,0xF0}; // undocumented but required magic unlock sequence
 	serial->write(unlock, 5);
 	serial->flush();
-	delay(10);*/
+	delay(10);
 	writeRegister(RSW, 0b0000000000010110);  // Return data content
 	delay(10);
 	writeRegister(RRATE, 0x08); // Return rate 0x06 = 10Hz (default), 0x08 = 50Hz, 0x09 = 100Hz
--- a/Firmware/LowLevel/lib/JY901_SERIAL/JY901.cpp
+++ b/Firmware/LowLevel/lib/JY901_SERIAL/JY901.cpp
@@ -10,26 +10,23 @@ CJY901 ::CJY901(SerialPIO *serial)
 	this->serial = serial;
 }
 
-void CJY901::begin(int baudrate) {
+void CJY901::begin(unsigned long baudrate) {
 	serial->begin(baudrate);
 	delay(1000);
-	uint8_t unlock[] = {0xFF,0xF0,0xF0,0xF0,0xF0};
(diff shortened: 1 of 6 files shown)


Firmware/LowLevel/lib/JY901_SERIAL/JY901.cpp
```diff
@@ -10,26 +10,23 @@ CJY901 ::CJY901(SerialPIO *serial)
 	this->serial = serial;
 }
 
-void CJY901::begin(int baudrate) {
+void CJY901::begin(unsigned long baudrate) {
 	serial->begin(baudrate);
 	delay(1000);
-	uint8_t unlock[] = {0xFF,0xF0,0xF0,0xF0,0xF0};
+	/*uint8_t unlock[] = {0xFF,0xF0,0xF0,0xF0,0xF0}; // undocumented / unknown?
```
Apehaenger: Couldn't find this unlock code in any documentation. Only in a doubtful Amazon comment. So I disabled it.

Firmware/LowLevel/lib/JY901_SERIAL/JY901.cpp
```diff
@@ -10,26 +10,23 @@ CJY901 ::CJY901(SerialPIO *serial)
...
 	delay(1000);
-	uint8_t unlock[] = {0xFF,0xF0,0xF0,0xF0,0xF0};
+	/*uint8_t unlock[] = {0xFF,0xF0,0xF0,0xF0,0xF0}; // undocumented / unknown?
 	serial->write(unlock, 5);
 	serial->flush();
+	delay(10);*/
+	writeRegister(RSW, 0b0000000000010110);  // Return data content
 	delay(10);
-	writeRegister(RSW, 0b0000000000010110);
+	writeRegister(RRATE, 0x08); // Return rate 0x06 = 10Hz (default), 0x08 = 50Hz, 0x09 = 100Hz
```
Apehaenger, line 22: IMU_CYCLETIME is 20ms, so I reduced WT's return rate to 50Hz

Firmware/LowLevel/lib/JY901_SERIAL/JY901.cpp
FadeFx: https://forum.arduino.cc/t/help-calibrating-witmotion-inclinometer-through-ttl-with-mega2560/941394/1
This is what my Google-fu gave me, not sure if it is related thou

ClemensElflein: it is needed I remember sniffing the configuration of the GUI app to see how they do it because just sending config didn't work

Apehaenger: Thanks for your comments!
After reading @FadeFx comment and link I immediately thought: "Could be an Auto-Baud detection sequence", but in-between tested it. It's not :-)
Will uncomment it again.

Apehaenger:
Deyan reported back yesterday, that this patch also works for his wt901, and that he's connected his WT901 back to j20 now.

PR #97, opened 2024-05-18
Shutdown ESCs when IDLE/EMERGENCY

olliewalsh:
Align IMU accelerometer axes.
Calculate pitch, roll, and overall tilt angles.

Set ESC shutdown pin to HIGH if:
  - shutdown_esc_max_pitch is set
  - not on a slope (pitch < shutdown_esc_max_pitch) and...
  - ROS is running and...
  - current mode is IDLE or emergency latched

Requires killswitch to be set to ADC2_HIGH in xesc app config

(Based on the work of @Pete_MI632 on discord)

### DIFF
--- a/Firmware/LowLevel/src/main.cpp
+++ b/Firmware/LowLevel/src/main.cpp
@@ -102,6 +102,7 @@ bool ROS_running = false;
 unsigned long charging_disabled_time = 0;
 
 float imu_temp[9];
+float pitch_angle = 0, roll_angle = 0, tilt_angle = 0;
 
 uint16_t ui_version = 0;                   // Last received UI firmware version
 uint8_t ui_topic_bitmask = Topic_set_leds; // UI subscription, default to Set_LEDs
@@ -395,6 +396,9 @@ void setup() {
     pinMode(LED_BUILTIN, OUTPUT);
     pinMode(PIN_ENABLE_CHARGE, OUTPUT);
     digitalWrite(PIN_ENABLE_CHARGE, HIGH);
+
+    pinMode(PIN_ESC_SHUTDOWN, OUTPUT);
+    digitalWrite(PIN_ESC_SHUTDOWN, LOW);
 
     gpio_init(PIN_RASPI_POWER);
     gpio_put(PIN_RASPI_POWER, true);
@@ -751,6 +755,12 @@ void loop() {
         sendMessage(&imu_message, sizeof(struct ll_imu));
 
         last_imu_millis = now;
+
+        // Update pitch, roll, tilt
+        pitch_angle = atan2f(imu_temp[0], imu_temp[2]) * 180.0f / M_PI;
+        roll_angle = atan2f(imu_temp[1], imu_temp[2]) * 180.0f / M_PI;
+        float accXY = sqrtf((imu_temp[0]*imu_temp[0]) + (imu_temp[1]*imu_temp[1]));
+        tilt_angle = atan2f(accXY, imu_temp[2]) * 180.0f / M_PI;
     }
 
     if (now - last_status_update_millis > STATUS_CYCLETIME) {
@@ -761,6 +771,25 @@ void loop() {
         status_message.v_charge =
                 (float) analogRead(PIN_ANALOG_CHARGE_VOLTAGE) * (3.3f / 4096.0f) * ((VIN_R1 + VIN_R2) / VIN_R2);
         status_message.charging_current = llhl_config.options.ignore_charging_current == OptionState::ON ? -1.0f : (float)analogRead(PIN_ANALOG_CHARGE_CURRENT) * (3.3f / 4096.0f) / (CURRENT_SENSE_GAIN * R_SHUNT);
+
+
+        // ESC power saving
+        bool shutdown_escs = false;
+        if (llhl_config.shutdown_esc_max_pitch != 0) {
+            bool emergency_or_idle = emergency_latch || HighLevelState::getMode(last_high_level_state.current_mode) == HighLevelState::Mode::IDLE;
+            bool on_slope = fabs(pitch_angle) > llhl_config.shutdown_esc_max_pitch;
+            if(ROS_running && emergency_or_idle && !on_slope) {
+                shutdown_escs = true;
+            }
+        }
+        if (shutdown_escs){
+            digitalWrite(PIN_ESC_SHUTDOWN, HIGH);
+            status_message.status_bitmask &= 0b11110111;
+        } else {
+            digitalWrite(PIN_ESC_SHUTDOWN, LOW);
+            status_message.status_bitmask |= 0b1000;
+        }
+
 
         status_message.status_bitmask = (status_message.status_bitmask & 0b11111011) | ((charging_allowed & 0b1) << 2);
         status_message.status_bitmask = (status_message.status_bitmask & 0b11011111) | ((sound_available & 0b1) << 5);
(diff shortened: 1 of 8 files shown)


Apehaenger:
~~Forget all I wrote... didn't recognized that it already got adapted and merge :-( What a pain.~~

Interesting PR! Planned to test it today, but after I reviewed it once more, I've some brain-blocker. Think it's due to my limited brain- resources, so please be so kind and help my rusty brain to get on track ;-)

> Set ESC shutdown pin to HIGH if:
>
>     * not on a slope (pitch < 15deg) and...
>
>     * ROS is running and...
> 
>     * current mode is IDLE

I don't get the logic behind this (as well as in the code).
Did I understood right that "ESC shutdown pin = high" is  "ESC Power status = off

If so: Why shutdown ESC when **not** on a slope?

What's the reason for shutting down the ESC's dependent on slope? Is it some kind of safety, but without triggering emergency?
If so, why not also triggering on roll(/tilt?) angle? Guess it's due to some IMU miss-understanding by me. I'm here when talking about  pitch/roll https://atadiat.com/en/e-towards-understanding-imu-basics-of-accelerometer-and-gyroscope-sensors/ section "*Computing the Pitch and Roll Angles Using Accelerometer*"

If "ESC shutdown" get triggered due to pitch(/roll), also the drive ESCs get shutdown right? Means we've a mower standing around without recognizing, because no emergency got triggered?

Please don't understand me wrong. I do like the PR but I miss somehow (parts of) the reasoning behind ;-)

rovo89:
> If so: Why shutdown ESC when not on a slope?

My understanding is that on a slope, the mower might roll off the docking station. So we need the motors to break actively, for which we need the ESC to be powered on.

rovo89:
Maybe you missed the point that this is for power saving and less heat while the mower is docked, not to stop the mowing motor in a potentially dangerous situation.

Apehaenger:
Wohoo! That's the part I missed ;-).
But wait ... with that new reason, it would (actively) break only if >15° pitch, isn't it?
Well the logic is clear then, but who the hack put his docking station on a >15° slope :-/ Because 15° is really a lot!! I guess the mower will roll out of his dock at >= 5°?!

Apehaenger:
> Maybe you missed the point that this is for power saving and less heat while the mower is docked, not to stop the mowing motor in a potentially dangerous situation.

Yes, I missed that fully!!

Anyway, I do like the idea to use the IMU for it and I  think we can/should use it for both reasons?

Apehaenger:
Like:
If docked (VCharge) && pitch > 5° enable ESC_Power (for active break) ||
If !docked && !idle && >15° => disable ESC for safety and trigger emergency?

rovo89:
> If !docked && !idle && >15° => disable ESC for safety and trigger emergency?

Can't really judge because my lawn is almost flat. I think some users have a steeper lawn though, and as long as the hall sensors don't raise an alarm, they probably don't want it to trigger emergency mode. And shutting down the wheel ESCs would be a bad idea in that case as well. Maybe it might make sense the other way around: If emergency is triggered, the mower ESC could be turned off, and also the wheel ESCs if it's *not* steep.

That said, I think it would be a completely different intention, different implementation and to be discussed with other affected users, so I would separate it from this PR.

I guess from your additional commits, the only thing to keep would be using the constant?

Apehaenger:
> > If !docked && !idle && >15° => disable ESC for safety and trigger emergency?
> 
> Can't really judge because my lawn is almost flat. I think some users have a steeper lawn though, and as long as the hall sensors don't raise an alarm, they probably don't want it to trigger emergency mode. And shutting down the wheel ESCs would be a bad idea in that case as well. Maybe it might make sense the other way around: If emergency is triggered, the mower ESC could be turned off, and also the wheel ESCs if it's _not_ steep.

Good point. Mine is also nearly 15° I guess, but didn't tested in real yet, that's why I didn't increased that value (with my wrong reasoned assumption). You're right, could be bad to shutdown the ESCs on a hill... even though it shouldn't trigger with a reasonable value like 20-25°, except one lift it from the back or side, in which case. Bla bla ... ;-)
 
> 
> That said, I think it would be a completely different intention, different implementation and to be discussed with other affected users, so I would separate it from this PR.

Okay, no problem. But >15° as roll-out undock protection look really very high to me.


> 
> I guess from your additional commits, the only thing to keep would be using the constant?

Take or drop whatever you (dis)like ;-)

FadeFx:
Guys, i am not sure, but i think you have a map of the yard in memory, you probably should also have stored surface angles, when taking the map. You also now exactly the possition of the mower at any time. Couldnt you just say turn shit off, if the angle of the mower differes a certain ammount from mapped angle?

Apehaenger:
> Guys, i am not sure, but i think you have a map of the yard in memory,

No, not in LowLevel ;-)

> you probably should also have stored surface angles,

No, don't think so. As far as I know our map is 2D

> when taking the map.

I'm sorry to say: "No not in LowLevel" ;-)

> You also now exactly the possition of the mower at any time.

Unfortunately not "*at any time*". We can't expect GNSS data in docking station

> Couldnt you just say turn shit off, if the angle of the mower differes a certain ammount from mapped angle?

No ;-). But luckily that's not the question anymore.

rovo89:
> you probably should also have stored surface angles, when taking the map

This is quite off-topic, but when recording the map, you only drive the outline of the area and obstacles. So there's no knowledge about any hills in the middle of the area. Besides that, the angles obviously depend on the mower orientation, so we'd have to be very sure about that one. I'd say it only makes sense to trigger emergency based on IMU for really extreme values (like: nose pointing down, mower upside down).

olliewalsh:
It will shutdown the ESCs when the mower status is IDLE (so docked or when stuck in emergency, although I think it will now stay mowing/PAUSED instead of failing to IDLE so need to update this logic), and not on a slope so it doesn't roll into a pond if it's stuck in emergency

rovo89:
@olliewalsh Any idea how to verify that it works properly? The only indicator I have right now is that the docking station consumes ~0.6W less when the mower is docked and fully charged. I bought an IR thermometer, but the measured temperature varies so much even under seemingly stable conditions that I don't trust it.

From d432b545fae765966e4e3cc7ab3296d08dee9738, only `kill_sw_mode` is relevant, right?

patrickzzz:
One sign that it works is, that you are able to turn the wheel when idle (if not on slope).. when picking up back on a higher angle, wheels should brake..

Btw. I checked behavior of stock mower: it has the brake activated in docking station in flat ground.. although for me it sounds reasonable not to use the break when not needed..

Firmware/LowLevel/src/main.cpp
```diff
@@ -762,6 +772,25 @@ void loop() {
                 (float) analogRead(PIN_ANALOG_CHARGE_VOLTAGE) * (3.3f / 4096.0f) * ((VIN_R1 + VIN_R2) / VIN_R2);
         status_message.charging_current = llhl_config.options.ignore_charging_current == OptionState::ON ? -1.0f : (float)analogRead(PIN_ANALOG_CHARGE_CURRENT) * (3.3f / 4096.0f) / (CURRENT_SENSE_GAIN * R_SHUNT);
 
+
+        // ESC power saving
+        bool shutdown_escs = false;
+        if (llhl_config.shutdown_esc_max_pitch != 0) {
```
olliewalsh, line 778: @Apehaenger was this the intended usage i.e only enable ESC power saving when the max pitch is set?

olliewalsh:
> i think you have a map of the yard in memory, you probably should also have stored surface angles

no, map is 2d

Firmware/LowLevel/src/main.cpp
Apehaenger: Sorry I missed the question. Yes exactly, it was indeed reserved for your PR ;-)

PR #104, opened 2024-06-13
Fix typo

rovo89:
~~Otherwise whenever *any* hall sensor triggers, it will trigger the STOP emergency bits after just 20 ms, instead of waiting 2500 ms (for tilt) or 100 ms (for lift).~~
Actually no, because only bits which are really in `emergency_read` are set. So worst case would be that some other hall sensor triggers the timer too early. It's still wrong though.

Already confirmed with @Apehaenger that this wasn't intended.


ClemensElflein:
Thank you for the PR, nice find

PR #106, opened 2024-08-23
Update main.cpp to add ADC filters to reduce noise on voltage and current sensor

Jahnkeanater:
Added an IIR filter with a weight of 90% to the battery voltage, charge voltage, and charging current to reduce the noise in the ADC measurements.


Jahnkeanater:
With the sample rate being only 10Hz I think a weight of 80% is probably better.

Jahnkeanater (commented):
Putting the power supply into PWM mode by pulling GPIO23 High should also help improve the ADCs.

#75 Removing this since it might actually make the noise worse.

PR #107, opened 2024-08-25
Improve battery handling

pelle-c:
The values given by the ADC are not very accurate. The introduced errors pose a risk for over-voltage on the battery, which in worst case can lead to an explosion.

This commit aims to mitigate the above by doing the following:

* Adding correction values in order for the reported values to match the real ones
* Adding IIR filter (from Jahnkeanater) to get more smooth readings
* Adding hysteresis for when to restart charging, to avoid charge/discharge toggling

### DIFF
--- a/Firmware/LowLevel/src/main.cpp
+++ b/Firmware/LowLevel/src/main.cpp
@@ -62,11 +62,38 @@ SerialPIO uiSerial(PIN_UI_TX, PIN_UI_RX, 250);
 #define R_SHUNT 0.003f
 #define CURRENT_SENSE_GAIN 100.0f
 
+// These are the absolute maximum voltage levels, for a 7S-pack with cells of 4.2V max the BATT_ABS_MAX will be 29.4V. Might be good 
+// with a safety margin here as well.
+// 7S parameters:
 #define BATT_ABS_MAX 28.7f
 #define BATT_ABS_Min 21.7f
+// 6S parameters:
+// #define BATT_ABS_MAX 25.2f
+// #define BATT_ABS_Min 19.2f
+
+// We really want to avoid going higher than BATT_ABS_MAX. All battery packs should have a bms that safeguards against overvoltage, 
+// but in worst case overvoltage can lead to explossions. 
+// Since the ADC will not give you very accurate readings, it might be worth measuring the values and try to correct the figures.
+// This can be done in the CORRECTION-defines.
+// Another step to avoid danger which can also prolong the life of the pack is not to charge it up to 100%. The threshold is
+// defined in BATT_FULL below. 
 
 #define BATT_FULL BATT_ABS_MAX - 0.3f
 #define BATT_EMPTY BATT_ABS_Min + 0.3f
+
+// Make some corrections to the ADC values. Assuming the error is linear the value from the adc is multiplied by the "FACTOR" and then "OFFSET" is added.
+#define ADC_BATTERY_CORRECTION_FACTOR 1.04f
+#define ADC_BATTERY_CORRECTION_OFFSET -0.2f
+#define ADC_CHARGE_CORRECTION_FACTOR 1.05f
+#define ADC_CHARGE_CORRECTION_OFFSET -0.4f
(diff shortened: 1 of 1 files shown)


pelle-c:
From other PRs I gather that ADC accuracy is somewhat a hot topic. When I began to look into this, I had +/- 2V fluctuations on voltage readings (mitigated by a capacitor on 3V3). But also after that, the readings could be up to 1V off. Which matters when it comes to charging batteries, avoiding overvoltage.
Is it me having bad ADCs in the pico, or are others also experiencing similar? Or in other words, is this PR meaningful at all?

karlranseierausrom:
I have not changed the hardware and observe high fluctuations of up to 3 V. This results in random mowing aborts, when the voltage seemd to drop for a very short time below the "battery low level". Therefore I think this PR is definently meaningful. For sure a hardware noise reduction would be better. 
Even the rain detector flashes (seen in the gui) often due to noise.
I looked at the values just in the gui. In my opinion it is not a typical noise, but more a repeating wave on the signal. I had no time to look at the values with an oscilloscope / spectrum analyser yet. If I should to that, would be nice if someone could tell me at which points I should measure.

pelle-c:
I posted a dump from my scope, measuring the 3V3, in openmower-hardware channel yesterday. 5V was ok.

Would be interesting to know if the correction values suite everyone or if they have to be tailored. Keeping in mind that perfection isn't required, but I guess at least a precision like +/- 0.1V would be nice.

Jahnkeanater:
I get about 0.2 volts of noise in the GUI. My biggest issue is the noise tripping the over current protection every single time my mower docks. I constantly have to flash the pico firmware to clear the OCP and get the mower to charge. Once it actually starts charging then it's typically 0.8-1.0 amps. I don't see the point of having 7 digits when we realistically only have about 1 decimal point of usable resolution. I'm curious if adding a capacitor to the ADC would help or if the pico just has really noisy ADCs.

Jahnkeanater:
I think I remember reading that the pico has an internal switching power supply that makes the ADC really noisy. "Driving high the SMPS mode pin (GPIO23), to force the power supply into PWM mode, can greatly reduce the inherent ripple of the SMPS at light load, and therefore the ripple on the ADC supply."

GP23 on the pico is pad TP4 on the bottom of the board. Also maybe adding a capacitor to the ADC Vref pin might help.

Jahnkeanater:
PXL_20240828_011020703 NIGHT~2

Here are all the GP23 locations on the top of the board.

Jahnkeanater:
I think just setting GPIO23 high in software will put the power supply in PWM mode.

#75 Never mind looks like PWM mode is slightly worse.

Also it seems like olliewalsh was working on a way for the mower to self calibrate when mowing. That seems better than hard coding a correction factor.

pelle-c:
I think this needs a bit more investigation. 
First, I believe that there is a hardware issue that doesn't apply to everyone, if it's a batch of bad picos or mb 0.13.2, I don't know. But I do know that my 3V3 was really ugly before I put a capacitor on it.
Secondly, agree that that the self calibrating PR (olliewalsh) looks nice (could be bootstrapped with a "good value" as well). Not sure that it's enough though, I had to not only add offset but also multiply to get decent values.
For me, the SMPS didn't make any noticeable difference. 
Will do some more testing when I find a moment.

Jahnkeanater:
My main board is 0.9.1. What voltage regulator are you using? I replaced the cheap Chinese one with the SMD components.
PXL_20220617_052830938~2

pelle-c:
I haven't modified anything on my 0.13.2 mb (apart from adding a capacitor on 3.3V). Believe the pico is generating the 3.3V.
Just tested to put a 2.5V refernce voltage on ADC_VREF, and suddenly the readings were good without corrections. The offset is about 0.1V and it fluctuates very little, maybe +/-0.01V.
Will do some more testing, but this does look promising at least on my mb.

pelle-c:
With the hardware mods below and the filter from Jahnkeanater, the values are good. I trimmed the pot to match a voltmeter when the battery was full and the error was about 0.1V when empty. So, this PR is not needed if the above is in place. 
om_v0_13_2_mods

PR #110, opened 2024-10-07
Enlargeable Config Packet

Apehaenger:
## Roadmap:

- [x] Decide for flexible config packet and struct
- [x] LowLevel (OpenMower Pico-FW)
  - [x] Implement packet handling
  - [x] config save in flash via LittleFS (rovo89) :heart_eyes:
  - [x] Implement config value handling
    - [x] Charge & Battery
    - [x] Hall sensor inputs
    - [x] Emergency handling
    - [x] Ignore_charging_current
    - [x] Misc (ignore charging current, ...)
- [x] HighLevel (open_mower_ros) please see https://github.com/ClemensElflein/open_mower_ros/pull/163
- [x] Test with old HL-FW
<sub>:hammer: = Currently working on</sub>

## Test:

Each side (LowLevel-Pico / HighLevel-Raspi) should be upgrade-able without any compatibility issues.
But for getting the new configuration possibilities, both sides need to be updated.

### LowLevel (Pico)
Firmware can be found within [Actions](https://github.com/ClemensElflein/OpenMower/actions).
Select the newest `Enlargeable Config Packet` workflow. Even if it has a red cross in front, it should contain the open-mower-pico-firmware package within the "Atrifacts" section (scroll down).

### HighLevel (Raspi)
Docker image can be pulled via `sudo podman pull ghcr.io/clemenselflein/open_mower_ros:pr-163`
(don't forget to change your /boot/openmower/openmower_version.txt to 'pr-163')

### Configuration workflow
Once HighLevel is up and detect a running LowLevel, it will send a "configuration wish-list" to LowLevel like:
```
Oct 30 17:37:08 Dolly openmower_launch[17587]: [ INFO] [1730306228.494246883]: Send ll_high_level_config packet 0x11
Oct 30 17:37:08 Dolly openmower_launch[17587]:          options{dfp_is_5v=0, background_sounds=0, ignore_charging_current=0},
Oct 30 17:37:08 Dolly openmower_launch[17587]:          v_charge_cutoff=-1.000000, i_charge_cutoff=-1.000000,
Oct 30 17:37:08 Dolly openmower_launch[17587]:          v_battery_cutoff=-1.000000, v_battery_empty=24.000000, v_battery_full=28.000000,
Oct 30 17:37:08 Dolly openmower_launch[17587]:          lift_period=65535, tilt_period=65535,
Oct 30 17:37:08 Dolly openmower_launch[17587]:          shutdown_esc_max_pitch=255,
Oct 30 17:37:08 Dolly openmower_launch[17587]:          language="en", volume=255
Oct 30 17:37:08 Dolly openmower_launch[17587]:          hall_configs="U, U, U, U, U, U, U, U, U, U"

```

LowLevel receive this "config packet", aligns it with his defaults values as well as his known hardware limits and answer with a config-response-packet containing the applied, adapted or default values, like:
```
Oct 30 17:37:08 Dolly openmower_launch[17587]: [ INFO] [1730306228.505676744]: Received ll_high_level_config packet 0x12
Oct 30 17:37:08 Dolly openmower_launch[17587]:          options{dfp_is_5v=0, background_sounds=0, ignore_charging_current=0},
Oct 30 17:37:08 Dolly openmower_launch[17587]:          v_charge_cutoff=30.000000, i_charge_cutoff=1.500000,
Oct 30 17:37:08 Dolly openmower_launch[17587]:          v_battery_cutoff=29.000000, v_battery_empty=24.000000, v_battery_full=28.000000,
Oct 30 17:37:08 Dolly openmower_launch[17587]:          lift_period=100, tilt_period=2500,
Oct 30 17:37:08 Dolly openmower_launch[17587]:          shutdown_esc_max_pitch=0,
Oct 30 17:37:08 Dolly openmower_launch[17587]:          language="en", volume=80
Oct 30 17:37:08 Dolly openmower_launch[17587]:          hall_configs="!L, !L, !S, !S, I, I, I, I, I, I"
```

### Configuration
Please check the sample config file regarding the possible new parameters, like [Battery & Charge](https://github.com/ClemensElflein/open_mower_ros/blob/38396194a4cc5d71de48d489fd35c19ca52f929a/config/mower_config.sh.example#L197-L210), [Hall/Emergency](https://github.com/ClemensElflein/open_mower_ros/blob/685b47fde61cd82336bcb76bc068ecc4f68ef705/config/mower_config.sh.example#L247-L282) or [Ignore Charging Current](https://github.com/ClemensElflein/open_mower_ros/blob/685b47fde61cd82336bcb76bc068ecc4f68ef705/config/mower_config.sh.example#L67-L70).


Firmware/LowLevel/include/debug.h
```diff
@@ -0,0 +1,47 @@
...
+// Define to stream debugging messages via USB
+//#define USB_DEBUG
+#define DEBUG_PREFIX "" // You may define a debug prefix string
+
+#ifdef USB_DEBUG
+#define DEBUG_SERIAL Serial
+// #define DfMiniMp3Debug DEBUG_SERIAL // Also output DFPlayer IN/OUT cmd data
+
+// Some bloody simple debug macros which superfluous '#ifdef USB_DEBUG' ...
+#define DEBUG_BEGIN(b)     \
```
rovo89: I guess you want to call this somewhere...?

Firmware/LowLevel/include/debug.h
```diff
@@ -0,0 +1,47 @@
...
+#ifndef _DEBUG_H_
+#define _DEBUG_H_
+
+// Define to stream debugging messages via USB
+//#define USB_DEBUG
+#define DEBUG_PREFIX "" // You may define a debug prefix string
+
+#ifdef USB_DEBUG
+#define DEBUG_SERIAL Serial
+// #define DfMiniMp3Debug DEBUG_SERIAL // Also output DFPlayer IN/OUT cmd data
```
rovo89: Seems to be a left-over from a different PR. If you want to keep this as a suggestion, I'd do add it in the other PR.

Firmware/LowLevel/include/debug.h
```diff
@@ -0,0 +1,47 @@
...
+    } while (0)
+#else
+#define DEBUG_BEGIN(b)
+#define DEBUG_PRINTF(fmt, ...)
+#endif
+
+#define PRINTF_BINARY_PATTERN_INT8 "%c%c%c%c%c%c%c%c"
+#define PRINTF_BYTE_TO_BINARY_INT8(i)                                                                               \
+    (((i)&0x80ll) ? '1' : '0'), (((i)&0x40ll) ? '1' : '0'), (((i)&0x20ll) ? '1' : '0'), (((i)&0x10ll) ? '1' : '0'), \
+        (((i)&0x08ll) ? '1' : '0'), (((i)&0x04ll) ? '1' : '0'), (((i)&0x02ll) ? '1' : '0'), (((i)&0x01ll) ? '1' : '0')
```
rovo89: Unused - if required somewhere else, it should be added there.

Firmware/LowLevel/src/datatypes.h
```diff
@@ -135,22 +137,62 @@ struct ll_ui_event {
...
-#define LL_HIGH_LEVEL_CONFIG_BIT_EMERGENCY_INVERSE 1 << 1  // Sample, for possible future usage, i.e. for SA-Type emergency
+#define LL_HIGH_LEVEL_CONFIG_BIT_BACKGROUND_SOUNDS 1 << 1  // Enable background sounds
 
 typedef char iso639_1[2]; // Two char ISO 639-1 language code
 
+enum class HallMode : uint8_t {
+    off = 0,
+    emergency,  // lift & tilt
+    stop,
+    pause
```
rovo89: I don't really know the implications of these modes, mind to explain (and/or add a comment)? I assume it has some to do with which of the timeouts to use?

Firmware/LowLevel/include/config.h
```diff
@@ -0,0 +1,19 @@
+#ifndef _CONFIG_H
+#define _CONFIG_H
+
+#define BATT_ABS_MAX 28.7f
+#define BATT_ABS_Min 21.7f
```
rovo89: - Shouldn't this be all upper-case?
- Rename all of these to `DEFAULT_BATTERY_ABS_MAX` etc.?
- What about `const float DEFAULT_BATTERY_ABS_MAX = 28.7f;` instead? I read that macros are frowned upon these days, especially for things that are clearly constants.

Firmware/LowLevel/src/datatypes.h
rovo89: Also, the enum values are basically constants, so I'd write them in upper-case.

Firmware/LowLevel/src/datatypes.h
```diff
@@ -135,22 +137,62 @@ struct ll_ui_event {
...
+    // ATTENTION: This is a flexible length struct. It is allowed to grow independently to HL without loosing compatibility,
+    //    but never change or restructure already published member, except you really know their consequences.
+
+    // uint8_t type; Just for illustration. Get set in wire buffer with type PACKET_ID_LL_HIGH_LEVEL_CONFIG_REQ or PACKET_ID_LL_HIGH_LEVEL_CONFIG_RSP
+
+    uint8_t config_bitmask = 0;                    // See LL_HIGH_LEVEL_CONFIG_BIT_*
+    uint16_t rain_threshold = 700;                 // If (stock CoverUI) rain value < rain_threshold then it rains. Expected to differ between C500, SA and SC types
+    float v_charge_cutoff = V_CHARGE_MAX;          // Protective max. charging voltage before charging get switched off
+    float i_charge_cutoff = I_CHARGE_MAX;          // Protective max. charging current before charging get switched off
+    float v_battery_cutoff = 29.0f;                // Protective max. battery voltage before charging get switched off
```
rovo89: Since you started adding constants, this could need one as well 😉

Firmware/LowLevel/src/datatypes.h
```diff
@@ -135,22 +137,62 @@ struct ll_ui_event {
...
+    float v_battery_empty = BATT_EMPTY;            // Empty battery voltage used for % calc of capacity
+    float v_battery_full = BATT_FULL;              // Full battery voltage used for % calc of capacity
+    uint16_t lift_period = LIFT_EMERGENCY_MILLIS;  // Period (ms) for both wheels to be lifted in order to count as emergency (0 disable, 0xFFFF do not change). This is to filter uneven ground
+    uint16_t tilt_period = TILT_EMERGENCY_MILLIS;  // Period (ms) for a single wheel to be lifted in order to count as emergency (0 disable, 0xFFFF do not change). This is to filter uneven ground
+    iso639_1 language = {'e', 'n'};                // ISO 639-1 (2-char) language code (en, de, ...)
+    int8_t volume = 80;                            // Volume (0-100%) feedback (if directly changed i.e. via CoverUI or WebApp)
```
rovo89: Maybe use `uint8_t`?

Firmware/LowLevel/include/debug.h
Apehaenger: Quite thanks for your review!

Yes, simply copied from Improve-Sound PR #80 because needed by nv_config and forgot to call. It's now in main.

BTW: How's pressing "Resolve conservation"? You as "Ack" I guess?

Firmware/LowLevel/include/debug.h
Apehaenger: Yip. Deleted. Will reintegrate with #80.

Firmware/LowLevel/include/debug.h
Apehaenger: You're right, also related to #80. Deleted.

Firmware/LowLevel/include/debug.h
rovo89: I don't have the button, so please press it for me :)

Firmware/LowLevel/src/datatypes.h
Apehaenger: > I don't really know the implications of these modes, mind to explain (and/or add a comment)? I assume it has some to do with which of the timeouts to use?

Yes, with both. Commented, but describing "lift" and "tilt" differences would be the wrong place there. That is, or should be done where the emergency get handled.

> Also, the enum values are basically constants, so I'd write them in upper-case.

I'm always unsure with this. If no style guide exists, I normally **try** follow "Google C++ Style Guide", but by enums i.e., they write that they should be named like const but NOT macros https://google.github.io/styleguide/cppguide.html#Enumerator_Names !
I never would write enum class member in UPPERCASE like macros but I also do NOT like google "k" prefix naming rule. Often tried, but looks and writes always ugly :-/

Firmware/LowLevel/src/main.cpp
```diff
@@ -367,7 +367,7 @@ void setup() {
     rp2040.idleOtherCore();
 
 #ifdef USB_DEBUG
-    DEBUG_SERIAL.begin(9600);
+    DEBUG_BEGIN(9600);
 #endif
```
rovo89: Nit-picking here, but I think you can just always call DEBUG_BEGIN, should be a no-op for non-debug (and I think you even had a comment about that in debug.h 😉).

Firmware/LowLevel/include/config.h
Apehaenger: Regarding "_Min" yes, changed. Copied from main and didn't validated.

Aren't all our macro values defaults? So no I wouldn't do so, also because I only moved them from main.cpp.
I always try to let them as they have been.

Regarding the const vs. macro. I'm undecided with this. A const take the whole footprint as a normal variable. A macro is simply inline expanded during compilation (what you know for sure). However, here I try the same: If the author preferred to use macros, I follow em, except the other make more sense. But mixing up without reason, I wouldn't do.

Firmware/LowLevel/src/datatypes.h
rovo89: I looked at the same guide... and agree that the "k" prefix looks strange. Not sure what else I'd use for real constants... I think upper-case. Oh right, [I did]( https://github.com/Apehaenger/OpenMower/commit/1c0cfacb801e5e21bd3c39f4b6be79a5c9e25708#diff-91326d111ce913c238d2694e904666d67eea719cc8f00bfb954e9f58ef421db6R122). So I guess for me, macros and constants share one style. And since the enum is basically just a namespace/prefix, I truly think that `HallLevel::ACTIVE_LOW` looks better than `HallLevel::activeLow`. Arduino seems to agree:
```c++
typedef enum {
  INPUT           = 0x0,
  OUTPUT          = 0x1,
  INPUT_PULLUP    = 0x2,
  INPUT_PULLDOWN  = 0x3,
  OUTPUT_2MA      = 0x4,
  OUTPUT_4MA      = 0x5,
  OUTPUT_8MA      = 0x6,
  OUTPUT_12MA     = 0x7,
} PinMode;
```

**But**: I already talked you out of so much of your original code already, and this is really an unimportant detail, so do what you like, I won't complain if you keep your style. 😆

Firmware/LowLevel/src/datatypes.h
Apehaenger: Yes good point :-)
This is in contrary to what I wrote before :-/

v_battery_cutoff was v_battery_max before, even it was not a max in my sense, but max as macro was redundant, so I took the static value.
With the integration of the new v_battery_empty/full values, _max couldn't stay as they are. So I rename that whole group as well as their macros... but didn't recognized that cutoff can become a macro now.
Well, changed now for the _cutoff value but not for rain_threshold... undecided if it make sense to give minor static values a macro if they're only used at one place :-/

Firmware/LowLevel/src/datatypes.h
Apehaenger: No, need int. It's an option :-)
Sound volume can be used as leading by ROS or leading by mower.
ROS send -1 if it doesn't modify the current setting

Firmware/LowLevel/include/config.h
rovo89: > Aren't all our macro values defaults?

No, `IMU_CYCLETIME` and many others are not. That's why I thought about indicating that these values aren't fixed, but can be changed later. How about simply renaming the file to "config_defaults.h"?

> A const take the whole footprint as a normal variable. A macro is simply inline expanded during compilation (what you know for sure).

Maybe the compiler also inlines `const`s and removes their footprint? Or maybe for a `static const`? Not sure.

But yeah, mixing up would be bad. All or nothing, so probably let it as-is.

Firmware/LowLevel/src/main.cpp
Apehaenger: Noo, not nit-picking, you're right. Brainlessly replaced DEBUG_SERIAL...
Now that you stub my nose to my own comment :-) I checked the macro to validate my comment :-).
I don't like all the `#ifdef debug` lines in main code, they disturb reading.
Removed but not committed. Will commit with next changes.

Firmware/LowLevel/src/datatypes.h
Apehaenger: I'm fully fine with your hint!
I don't have community-wise dev experience, and I really like having all that rules. But I miss that routine, I still need to think how was what written?

Regarding the enum: You're right! Clemens also write them in UPPERCASE. In the same file! So, for sure I follow that style.

Firmware/LowLevel/include/config.h
Apehaenger: Yea "config_defaults.h" is fine. I'm bad with finding good names at all. Doesn't matter if function, method, class or file name :-(

rovo89:
Seems like you're not using anything from debug.h now, except the initialization. 🤔

rovo89:
IGNORE_CHARGING_CURRENT could also become a setting I guess. Maybe we can have only a single build in the future?

rovo89:
> Seems like you're not using anything from debug.h now, except the initialization. 🤔

Ah, in main.cpp, there are many places with `#ifdef USB_DEBUG DEBUG_SERIAL.println(...)`. Did you mean to replace those? But that would be unrelated to this PR. So in the interest of keeping this PR focused on the extended config, I'd suggest to undo the `DEBUG_BEGIN` change and remove debug.h. If you feel that it simplifies stuff, it can be introduced in another PR.

Firmware/LowLevel/include/config_defaults.h
```diff
@@ -0,0 +1,26 @@
+#ifndef _CONFIG_H
```
rovo89: `_CONFIG_DEFAULTS_H`?

Firmware/LowLevel/include/config_defaults.h
```diff
@@ -0,0 +1,26 @@
...
+#define BATT_EMPTY BATT_ABS_MIN + 0.3f
+#define BATT_FULL BATT_ABS_MAX - 0.3f
+
+#define V_BATT_CUTOFF 29.0f
+
+#define TILT_EMERGENCY_MILLIS 2500  // Time for a single wheel to be lifted in order to count as emergency (0 disable). This is to filter uneven ground.
+#define LIFT_EMERGENCY_MILLIS 100   // Time for both wheels to be lifted in order to count as emergency (0 disable). This is to filter uneven ground.
+
+#define V_CHARGE_MAX 30.0f  // Default YF-C500 max. charge voltage
+#define I_CHARGE_MAX 1.5f   // Default YF-C500 max. charge current
```
rovo89: `..._CHARGE_CUTOFF`, like in the struct? Doesn't seem to be used anywhere else.

Firmware/LowLevel/src/datatypes.h
```diff
@@ -135,22 +137,62 @@ struct ll_ui_event {
...
 typedef char iso639_1[2]; // Two char ISO 639-1 language code
 
+enum class HallMode : uint8_t {
+    OFF = 0,
+    EMERGENCY,  // Triggers emergency with lift & tilt functionality
+    STOP,       // Stop mower
+    PAUSE       // Pause the mower (not yet implemented in ROS)
+};
+
+// FIXME: Decide later which is more comfortable, activeLow = 0 | 1
```
rovo89: When I add a `FIXME` comment, it usually means stuff that I want to address before merging - is there still anything to decide right now?

Firmware/LowLevel/include/config_defaults.h
rovo89: Sorry if I had asked this before, don't remember if I talked out loud or just thought about it 😆:
Do we actually need/want to have all these macros in one include separated from the structure definition?

I've just sent you a [PR](https://github.com/Apehaenger/OpenMower/pull/2) to actually use the runtime config instead of these macros in main.cpp, and that makes these macros single-use. So the only purpose that comes to my mind is to make them easy to change for a custom build. But isn't that the opposite of what we're just trying to achieve? Aren't people supposed to set the correct values in their runtime config instead of messing with the *defaults*?

Again, I have [prepared something](https://github.com/rovo89/OpenMower/compare/use_config...rovo89:OpenMower:no_defaults_include?expand=1) to show what it would look like if the defaults were inlined into the structure definition. It still compiles, and from my point of view it simplifies the code without taking any freedom away.

Apehaenger:
> Seems like you're not using anything from debug.h now, except the initialization. 🤔

No, not anymore, since you deleted nv_config :-)

> IGNORE_CHARGING_CURRENT could also become a setting I guess. Maybe we can have only a single build in the future?

I already had that discussion in a similar way a couple of weeks ago. I liked to move it to checkShouldCharge() and don't check for the current, but got stopped with something like "better no values" than "wrong values" (even if the relevant ppl know their hardware failure)... But wait...  or do you mean those could configure/send `i_charge_cutoff = 0` and we disable the cutoff logic based on the current?
No, I wouldn't do that. All ppl. could disable the charge logic by this. Yes, all people can also flash *_IGNORE_CHARGE_CURRENT, that's true :thinking: . Please ask @ClemensElflein for this, I wouldn't do that, simply because the name kindly sound dangerous and the normal healthy use will not install that version.



> > Seems like you're not using anything from debug.h now, except the initialization. 🤔
> 
> Ah, in main.cpp, there are many places with `#ifdef USB_DEBUG DEBUG_SERIAL.println(...)`. Did you mean to replace those? But that would be unrelated to this PR. So in the interest of keeping this PR focused on the extended config, I'd suggest to undo the `DEBUG_BEGIN` change and remove debug.h. If you feel that it simplifies stuff, it can be introduced in another PR.

Will do with the next commit ;-)

Firmware/LowLevel/include/config_defaults.h
Apehaenger: Oh yes! Changed.

rovo89:
> But wait... or do you mean those could configure/send `i_charge_cutoff = 0` and we disable the cutoff logic based on the current?

I hadn't checked in much detail - just thought that we could try and reduce the number of build variants. 😉

IIUC, `IGNORE_CHARGING_CURRENT` acts as if the charging current sensor always reported -1.0A. That affects the value reported to ROS, the status LED, but also `checkShouldCharge()`. Since -1.0 will be lower than any configured `i_charge_cutoff`, this will never prevent charging.

I have no personal interest here. Again, I'm just looking at how this PR can be utilized, and since you asked for more settings, that was one that came to my mind. (Another one would be settings for for #97, but since that isn't merged yet, we can also add them there.)

From risk perspective, I guess it's no different from setting `i_charge_cutoff` to a huge value (or even `Inf`)?!? Except that `IGNORE_CHARGING_CURRENT` also makes sure that no random values are sent to ROS, which might be confusing in the UI. I think that was the "better no values than wrong values" part.

Firmware/LowLevel/include/config_defaults.h
Apehaenger: Yes, but not yet sure how to handle/name.
Need to clear that up when implementing the handling of the received values.
Naming of these look somehow wildly grown in past. There where some _MAX also _ABS_MAX, _EMPTY, _FULL and also some static ones before. 
I started by integrating the CUTOFF ones to make clear that those switch charging off, but I need to recheck and rethink about those because I don't wanna integrate them all in ROS. I like to use only the sensor specific ones from ROS which are Critical_Low, min, max and critical high. Initially I thought to use the critical_height @ ROS for CUTOFF @ LL and the min/max ones for full/empty... but than occurred the _ABS_ ones.
Might happen that I rename the _CUTOFFs to _ABS... later, but will check that in detail when implementing the config value handling.

Firmware/LowLevel/src/datatypes.h
Apehaenger: Me too, but in my understanding something which needs to be fixed before merging (if technical possible).

And "yes", a loot. See my top post as well as this PR is in "Draft"

Apehaenger:
> 
> I have no personal interest here. Again, I'm just looking at how this PR can be utilized, and since you asked for more settings, that was one that came to my mind. (Another one would be settings for for #97, but since that isn't merged yet, we can also add them there.)

Fully clear and I'm grateful for getting such a detailed review and good hints!!
Regarding #97: Now that we try to have nothing unrelated in this PR, lets go on this way, even if it would be more clever to add SHUTDOWN_ESC_MAX_PITCH now, because the one who add that, has also to decide for the num of hall spares :-/

>
> From risk perspective, I guess it's no different from setting `i_charge_cutoff` to a huge value (or even `Inf`)?!? Except that `IGNORE_CHARGING_CURRENT` also makes sure that no random values are sent to ROS, which might be confusing in the UI. I think that was the "better no values than wrong values" part.

Yes, and that time, my reasoning for moving the IGNORE implementation to checkShouldCharge() was the opposite, because a couple of ppl. use this IGNORE thingy for an increased battery charging over the fixed 1.5A and don't see if they charge with 2A or 3A. But that should be solved soon ;-)

Firmware/LowLevel/include/config_defaults.h
rovo89: Did you see my other comment above? https://github.com/ClemensElflein/OpenMower/pull/110#discussion_r1797773681
I already sent a PR for "when implementing the handling of the received values", and I'm wondering whether we need those macros at all (as outlined in that comment).

Firmware/LowLevel/src/datatypes.h
rovo89: I was only referring to decisions about this single line 😆 IIUC, it's just about whether to encode ACTIVE_LOW as value 0 or 1? Shouldn't matter, but 0 makes more sense to me.

Firmware/LowLevel/include/config_defaults.h
Apehaenger: > Sorry if I had asked this before, don't remember if I talked out loud or just thought about it 😆: Do we actually need/want to have all these macros in one include separated from the structure definition?
> 
Well, it was one of your review hints that you picked static values out of the struct and asked if they shouldn't go into a macro at a common place :-/
I don't care because it's pretty simple with VSCode to hover over a macro and see their value.

BTW: Regarding const vs. macro (because we talked about it): Did a quick check while integrating your [LittleFS PR](https://github.com/Apehaenger/OpenMower/pull/1). The filename const took 12 Bytes RAM as well as 32 Byte Flash **more** in comparison to the macro. That's a pretty clear const- no-go for embedded projects (as long as none of the other const benefits are needed)

> I've just sent you a [PR](https://github.com/Apehaenger/OpenMower/pull/2) to actually use the runtime config instead of these macros in main.cpp, and that makes these macros single-use. So the only purpose that comes to my mind is to make them easy to change for a custom build. But isn't that the opposite of what we're just trying to achieve? Aren't people supposed to set the correct values in their runtime config instead of messing with the _defaults_?
> 

> Again, I have [prepared something](https://github.com/rovo89/OpenMower/compare/use_config...rovo89:OpenMower:no_defaults_include?expand=1) to show what it would look like if the defaults were inlined into the structure definition. It still compiles, and from my point of view it simplifies the code without taking any freedom away.

Quite thanks for the PRs! I didn't recognized them in time, because I was working on the other (ROS) side.
Will merge once I'm done there, because I expect some conflicts now.

Firmware/LowLevel/include/config_defaults.h
Apehaenger: No, haven't seen it before, but now ;-)

Firmware/LowLevel/include/config_defaults.h
rovo89: > Well, it was one of your review hints that you picked static values out of the struct and asked if they shouldn't go into a macro at a common place :-/

Yes... I feel bad about this now 😔 Back then, it was for consistency reasons because you had most defaults in macros already. But for code simplicity (even without tools) and the arguments mentioned above, I would now strongly prefer to get rid of the include completely. Sorry again!

> The filename const took 12 Bytes RAM as well as 32 Byte Flash more in comparison to the macro.

Did you also try it with a `static const`?

rovo89: > > The filename const took 12 Bytes RAM as well as 32 Byte Flash more in comparison to the macro.
> 
> Did you also try it with a `static const`?

I could reproduce the memory difference - until I realized that the overhead really came from using `String`:

```
const String CONFIG_FILENAME = "/openmower.cfg";
RAM:   [=         ]   5.9% (used 15496 bytes from 262144 bytes)
Flash: [          ]   4.9% (used 103136 bytes from 2093056 bytes)

static const String CONFIG_FILENAME = "/openmower.cfg";
RAM:   [=         ]   5.9% (used 15496 bytes from 262144 bytes)
Flash: [          ]   4.9% (used 103136 bytes from 2093056 bytes)

-----

#define CONFIG_FILENAME "/openmower.cfg"
RAM:   [=         ]   5.9% (used 15484 bytes from 262144 bytes)
Flash: [          ]   4.9% (used 103104 bytes from 2093056 bytes)

const char *CONFIG_FILENAME = "/openmower.cfg";
RAM:   [=         ]   5.9% (used 15484 bytes from 262144 bytes)
Flash: [          ]   4.9% (used 103104 bytes from 2093056 bytes)

static const char *CONFIG_FILENAME = "/openmower.cfg";
RAM:   [=         ]   5.9% (used 15484 bytes from 262144 bytes)
Flash: [          ]   4.9% (used 103104 bytes from 2093056 bytes)
```

So just change it to `const char *CONFIG_FILENAME`. 😉

Firmware/LowLevel/src/datatypes.h
Apehaenger: I also referred about that single line :-)

No wouldn't matter much, but like to decide later because even if ACTIVE_LOW as 0 sound logic, it isn't, because ACTIVE_LOW also means "inverted". Have had the enum at first as "normal" and "inverted" but within the code when working with the GPIOs, "normal" is somehow undefined. That's why I switched to the more technical word "Active Low" and "Active High" because it's more precise and normally clear for a developer.

Later on, within mower_config, I think we shouldn't use active_* because to technical for the normal user. The normal user knows better with "inverted" or "normal" (non-inverted).

And lastly: later when building the emergency logic, it should also look straight forward and not popup questions like "Why does he invert that again when it's already active_high?".
I don't waste thinking about it ATM, will decide later when I'm clear with the code, and changing this enums later is the simplest ;-)

Firmware/LowLevel/include/config_defaults.h
Apehaenger: Ahh good to know :+1: 

You're kidding, isn't it? :-)
It would be okay for a new project, but imagine how a single `const char *whatever` would look between all the other `#define whatever` lines.

rovo89: I wasn't kidding 😆 But yeah, `#define` is of course also fine.

The bigger difference would be getting rid of config_defaults.h, as I think that will simplify the code quite a bit. And again sorry for making you go the other way first. :/

Apehaenger: :laughing: 
Don't worry, config_defaults will get removed. I'll merge your PR. But I stuck currently in LL/HL testing, Bitfield defaults and the like... will get that ready before ;-)

I'll go to bed now, but probably you know?
I already added the constructor
```
struct HallConfig {
  HallConfig() : mode(HallMode::UNDEFINED), level(HallLevel::ACTIVE_LOW) {};
  HallMode mode : 3;
  HallLevel level : 1;
} __attribute__((packed));
```
Thought the constructor get called with empty default initializer
```HallConfig hall_configs[MAX_HALL_INPUTS] = {};  // Set all to UNDEFINED```
but I was wrong :-1: 
There must be a more clever way than writing them 10 times as initializer or adding a init loop.

Comment (rovo89) [reply Firmware/LowLevel/include/config_defaults.h]: Strange, I would have thought the same, and so does ChatGPT: https://chatgpt.com/share/670c4845-58fc-8001-bd96-d38897031054
I know that it doesn't care too much about truth, but it came with the nice idea to add logging to the standard constructor to find out whether it was called.

Comment (Apehaenger) [reply Firmware/LowLevel/include/config_defaults.h]: Ahh cool, she might be right. I tested only with aggregate initializer, and she said that zero initialization will be chosen then. Will try without it 👍.
Quite thanks 👍

Already googled a couple of minutes yesterday, but your ChatGPTs answer is the most matching and best describing one! I do need to test/ChatGPT some time 🤔

**Comment by Apehaenger** [comment]
Time for an intermediate result.
Tested the protocol together with ROS and made some alignments, nothing more except that I also implemented the last discussed stuff as well as that I added `shutdown_esc_max_pitch` config member. It get merged earlier or later anyway ;-)

Next, I'll check if LittleFS is working as expected.

ROS side's PR https://github.com/ClemensElflein/open_mower_ros/pull/163

File: Firmware/LowLevel/src/main.cpp [hunk 12]
```diff
@@ -372,9 +369,7 @@ void setup() {
     //  Therefore, we pause the other core until setup() was a success
     rp2040.idleOtherCore();
 
-#ifdef USB_DEBUG
-    DEBUG_SERIAL.begin(9600);
-#endif
+    DEBUG_BEGIN(9600);
```
rovo89: I think that's no longer defined now...?

Apehaenger: Sure, that's how Clemens made the uart debugging. My one was only an extension so that I do not write all the #ifdef's and that stuff :wink:

rovo89: Sure, but wasn't `DEBUG_BEGIN` only defined in `debug.h`, which is now gone?
Sorry, I didn't pull your latest branch and try to compile it, therefore it's just a review from looking at the code. I've been trying to train an AI model for the past few days, still not working... :/

Apehaenger: That's how it look now:
```
#ifdef USB_DEBUG
    DEBUG_SERIAL.begin(9600);
#endif
```
Look like you're in the wrong files

Firmware/LowLevel/src/datatypes.h
```diff
@@ -135,22 +136,66 @@ struct ll_ui_event {
...
+    float v_charge_cutoff = 30.0f;         // Protective max. charging voltage before charging get switched off (NAN = unknown)
+    float i_charge_cutoff = 1.5f;          // Protective max. charging current before charging get switched off (NAN = unknown)
+    float v_battery_cutoff = 29.0f;        // Protective max. battery voltage before charging get switched off (NAN = unknown)
+    float v_battery_empty = 21.7f + 0.3f;  // Empty battery voltage used for % calc of capacity (NAN = unknown)
+    float v_battery_full = 28.7f - 0.3f;   // Full battery voltage used for % calc of capacity (NAN = unknown)
+    uint16_t lift_period = 100;            // Period (ms) for both wheels to be lifted in order to count as emergency (0 = disable, 0xFFFF = unknown). This is to filter uneven ground
+    uint16_t tilt_period = 2500;           // Period (ms) for a single wheel to be lifted in order to count as emergency (0 = disable, 0xFFFF = unknown). This is to filter uneven ground
+    float shutdown_esc_max_pitch = 15.0f;  // Do not shutdown ESCs if absolute pitch angle is greater than this (to be implemented)
```
rovo89: In https://github.com/ClemensElflein/OpenMower/pull/97, the whole feature is behind an #ifdef. If we want to keep it optional in the feature without having to build separate variants for it, how about adding a boolean toggle for it here?

Firmware/LowLevel/src/main.cpp
```diff
@@ -547,13 +548,56 @@ void onUIPacketReceived(const uint8_t *buffer, size_t size) {
...
+    size_t payload_size = min(sizeof(ll_high_level_config), size - 2);  // exclude crc
+
+    // Use a temporary config for easier sanity adaption and copy our live config, which has at least reasonable defaults.
+    // The live config copy ensures that we've reasonable values for the case that HL config struct is older (smaller) than ours.
+    auto hl_config = llhl_config;
+
+    // Copy payload to temporary config
+    memcpy(&hl_config, buffer, payload_size);
+
+    // Takeover of HL leading config values
```
rovo89: Mind explaining what all this logic for "HL leading" is about? I thought that HL would always send the full config and we take over everything (expect on version mismatch, but I think we can ignore that for a second). You rather want to be able to send a delta config, which keeps "undefined" fields as they are? Is that really necessary and worth the complexity? I think you mentioned volume which can be set on the mower, but is there anything else that would be controlled from LL?

Firmware/LowLevel/src/datatypes.h
Apehaenger: Oh didn't checked in that detail.
Would prefer kind of "undefined = NAN" in which case shutdown functionality could be skipped?

Firmware/LowLevel/src/main.cpp
Apehaenger: Yes sure, will try, might be that it sounds somehow confusing :smile: 

At the moment we need to maintain/align a couple of LL configs settings also in HL. In my point of view this is error prone and redundant, as well as HL should not care much about LL config values/defaults.

I know, this sounds contrary to what I plan to implement. But we do have two use case scenarios.
\>90?% will go with the default LL settings whereas some others like to use adapted ones.

My target is that HL only inform LL about the adapted ones, whereas LL inform about the used (sanity aligned/default) ones.

Yesterday I build parts of that and I need to say that it also look somehow ugly as well as confusing to me :disappointed: ... later in bed I recognized that I probably only need to change the config- triggering side from LL to HL which is much more logic as well as it save some stupid looking assignments.

Now that the protocol is done in general and the communication work, I'll try a little bit with one value and look how it could be made more straightforward. Will give you an code example when I got it together in a more clear way.

rovo89: Most of this is about the defaults though? Apart from volume, is there any other setting that is changeable from the mower side?

Apehaenger: Yes. In my vision also all other values like charge, battery, timeouts, ... if not separately defined in HL (mower_config).

I'm almost done with my implementation. Let's simply wait one or two more days if I'm on the wrong track. If not, code says more than 1000 words :laughing: 

BTW: Think I know how to get rid of `IGNORE_CHARGING_CURRENT`. How's about giving it a bit in https://github.com/Apehaenger/OpenMower/blob/37f2ac6d2f30e52d7e6433d08c7101fa9e7e4b48/Firmware/LowLevel/src/datatypes.h#L177 ?
But like to give it a more clear name then. Something like `WRONG_IJL` or however the current-sense-chip got called.

rovo89: > Yes. In my vision also all other values like charge, battery, timeouts, ... if not separately defined in HL (mower_config).

But those wouldn't be changed "live"/at runtime by pressing some button on the mower, right?
I think you're saying that you don't want to maintain the defaults in two different places, which I think is a good point. I wanted to make up my mind about this, and therefore wanted to know how "dynamic" the config is on LL side.

> BTW: Think I know how to get rid of `IGNORE_CHARGING_CURRENT`. How's about giving it a bit in https://github.com/Apehaenger/OpenMower/blob/37f2ac6d2f30e52d7e6433d08c7101fa9e7e4b48/Firmware/LowLevel/src/datatypes.h#L177 ?
> But like to give it a more clear name then. Something like `WRONG_IJL` or however the current-sense-chip got called.

Using the bitmask sounds good. Maybe bits can be accessed more easily with an enum like you did with the halls? But for the name... your suggestion is very cryptic to me. Maybe `BROKEN_CHARGING_CURRENT_SENSOR`?

rovo89: I think I had a thought yesterday to have a "mask" to define which bytes/fields of the HL config shall be copied to LL. So instead of the memcpy, only copy the overridden bytes. Instead of saying what "undefined" means for each field and having at least one if-then-else per field.

Also not sure how flash would be handled, should it also store only the overridden fields?

But my mind is a bit blocked right now. If something else comes to mind, I'll share it with you (especially if it's easier), otherwise I'll see what you came up with. 😉

Apehaenger: > > Yes. In my vision also all other values like charge, battery, timeouts, ... if not separately defined in HL (mower_config).
> 
> But those wouldn't be changed "live"/at runtime by pressing some button on the mower, right?

No. This is only for volume (and in past also language change)... and I do not plan at the moment, that a volume change triggers a new config exchange. It only needs to be saved in flash so that a new power-on start with the right volume (and language).

The ones I'm mainly interested in, are only the default (or saved) ones and those get only exchanged when starting up or LowLevel got re-flashed.

> I think you're saying that you don't want to maintain the defaults in two different places, which I think is a good point. I wanted to make up my mind about this, and therefore wanted to know how "dynamic" the config is on LL side.
> 

LL has it's C500 "default" values from firmware, as well as the ones which are limited by hardware (limited by MOSFET or diode).
The "default" ones can be overwritten by HL for those people who use i.e. higher charging currents or different battery packs, not to forget the halls.
Those all get saved into flash (as you know), so that LL always boot up with some reasonable values.

I like to get it that way that HL doesn't need to know any of the LL values. LL shall inform HL about them.
And I also like to get it that way, that HL doesn't need to know about possible max ones. LL will change a possible to-large value on receive and send back in the return packet.

> > BTW: Think I know how to get rid of `IGNORE_CHARGING_CURRENT`. How's about giving it a bit in https://github.com/Apehaenger/OpenMower/blob/37f2ac6d2f30e52d7e6433d08c7101fa9e7e4b48/Firmware/LowLevel/src/datatypes.h#L177 ?
> > But like to give it a more clear name then. Something like `WRONG_IJL` or however the current-sense-chip got called.
> 
> Using the bitmask sounds good. Maybe bits can be accessed more easily with an enum like you did with the halls? But for the name... your suggestion is very cryptic to me. Maybe `BROKEN_CHARGING_CURRENT_SENSOR`?

Enum... hmm :thinking: , in the case of the halls it's really an "either or" so enum fits, but the bitmask is more a switch for a couple of things. Will take a look... probably a bitfield which would be more modern but should fit?!

Alternative `IGNORE_CHARGING_CURRENT` name:
Just checked, it's name is TSC101 :thinking: ... really not better :laughing: 
But yeees, much better! But it's not broken, simply wrong assembled. So `WRONG_CHARGING_CURRENT_SENSOR`? :+1:

rovo89: I think I understand your motivation. In short, we want to use the defaults and overlay them only with those values from HL which have been explicitly set. And the LL->HL config packet is mainly for displaying the actually used values.

Will take a step back and think about it.

> Enum... hmm 🤔

It was new to me, but you used [this](https://github.com/Apehaenger/OpenMower/blob/37f2ac6d2f30e52d7e6433d08c7101fa9e7e4b48/Firmware/LowLevel/src/datatypes.h#L161-L164). So I was wondering if you could have a list of `boolean xyz : 1` members? No idea whether that's valid syntax.

Apehaenger: > I think I understand your motivation. In short, we want to use the defaults and overlay them only with those values from HL which have been explicitly set. And the LL->HL config packet is mainly for displaying the actually used values.

Correctly :+1: .
We need them not only for minor stuff like sensor view, but also for major stuff like battery charge level which decide when to go back home.
Beside those, the LL->HL config packet is also to informing HL about probably to large values (requested before from HL) due to hardware limits.

> 
> Will take a step back and think about it.
> 
> > Enum... hmm 🤔
> 
> It was new to me, but you used [this](https://github.com/Apehaenger/OpenMower/blob/37f2ac6d2f30e52d7e6433d08c7101fa9e7e4b48/Firmware/LowLevel/src/datatypes.h#L161-L164). So I was wondering if you could have a list of `boolean xyz : 1` members? No idea whether that's valid syntax.

Yes :+1, something like:
```
struct Options {
    bool dfp_is_5v : 1;
    bool background_sound : 1;
    bool wrong_charging_current_sensor: 1;
}
```
But I'm also unsure at the moment how the code would look. But as far as I know, bitfields got implemented exactly for such a case like ours. Insofar, probably good hint to push me away from the good old binary masking to some more modern stuff :laughing:

rovo89: My dominant thought is: We should have considered this when we talked all the time about how to define the struct, do the memcpy, automatically apply/keep defaults etc. while keeping compatibility between versions. The solution we finally agreed on feels elegant to me - but when HL is the source of truth, not when you just want to send a small subset of the fields. 😕 Especially the little magic with the memcpy isn't used at all.

Would it really be so bad to set defaults in the firmware that are safe for "all" mowers, and define either the same or more mower-specific values in HL? That would avoid uncertainties about priorities (is HL, LL, flash leading?), simplify the code (despite a bit more config), with the additional advantage that updated defaults could be included in a new image, which is much more likely to get applied than a new firmware. The part with the LL->HL communication still makes sense to me to see the actually used values after validation.

Apehaenger: I'm sorry, I though it's clear because it's how the config packet is live since months :innocent: 
Beside that, I asked to discuss how that existing packet can be made enlargeable without tackling with LL/HL compatibility (my old packet-trailer idea).
And to be clear, I'm still exited about the result of that discussion because we do now have a perfect flexible and compatible solution.

Don't worry about the packet/struct. It's pretty clear, straightforward, and since I changed now the packet- trigger- side from LL to HL (still local here), it became once more simpler (at least in my point of view :laughing: ).

But let's wait if I succeed, and then take a look. If it turns out that it's bad and to complicated, then it simply get trashed ;-)

Will give an intermediate commit once I do have two config values ready, so that you can take a view how it look

Apehaenger:
Please see here https://github.com/ClemensElflein/open_mower_ros/pull/163#issuecomment-2423292585 about state/description

Apehaenger:
In my opinion, all is done.
Did a couple of tests with old-LL/new-HL, new-LL/old-HL, new-LL/new-HL on C500 as well as SA600? mower.
Haven't found any issue.

Would love to see a generic review (also from @ClemensElflein) before asking for tester @ discord.

karlranseierausrom:
May I already test? Is there an easy way to change the firmware version? I would like to check voltage and current settings to fit to my bigger battery.

Apehaenger:
> May I already test? Is there an easy way to change the firmware version? I would like to check voltage and current settings to fit to my bigger battery.

Sure, testing is always welcome :innocent: 

Pico-FW is here: https://github.com/ClemensElflein/OpenMower/actions/runs/11534325476/artifacts/2107987830
(if you don't know how to flash it, pls. contact me on discord)

ROS Image via: `sudo podman pull ghcr.io/clemenselflein/open_mower_ros:pr-163`
(don't forget to change your `/boot/openmower/openmower_version.txt` to `pr-163`)

Please check the sample config file regarding the required parameter: https://github.com/ClemensElflein/open_mower_ros/blob/38396194a4cc5d71de48d489fd35c19ca52f929a/config/mower_config.sh.example#L197-L210

You may watch the `ll_high_level_config` word in the openmower-log to follow up the configuration.

ClemensElflein (approved):
LGTM

Firmware/LowLevel/src/datatypes.h
```diff
@@ -34,21 +34,17 @@ enum HighLevelMode {
...
 };
 
-#define LL_EMERGENCY_BIT_LATCH 0b00000001
-#define LL_EMERGENCY_BIT_HALL1 0b00001000 // Lift1
-#define LL_EMERGENCY_BIT_HALL2 0b00010000 // Lift2
-#define LL_EMERGENCY_BIT_HALL3 0b00000010 // Stop1
-#define LL_EMERGENCY_BIT_HALL4 0b00000100 // Stop2
+#define LL_EMERGENCY_BIT_LATCH (1 << 0)
+#define LL_EMERGENCY_BIT_LIFT (1 << 1)  // Lift (or tilt)
+#define LL_EMERGENCY_BIT_STOP (1 << 2)  // Stop
```
rovo89: Maybe align the values?

Firmware/LowLevel/src/datatypes.h
```diff
@@ -34,21 +34,17 @@ enum HighLevelMode {
...
-#define LIFT2_IS_INVERTED 0
+// Need the old emergency_bitmask definition because CoverUI send it
+#define LL_EMERGENCY_BIT_CU_LIFT  0b00001000  // LIFT | LIFTX (up to CoverUI FW 2.0x)
+#define LL_EMERGENCY_BIT_CU_BUMP  0b00010000  // LBUMP | RBUMP (up to CoverUI FW 2.0x)
+#define LL_EMERGENCY_BIT_CU_STOP1 0b00000010  // Stop1
+#define LL_EMERGENCY_BIT_CU_STOP2 0b00000100  // Stop2
+#define LL_EMERGENCY_BIT_CU_LIFTX 0b00100000  // CoverUI-LIFTX (as of CoverUI FW 2.1x)
+#define LL_EMERGENCY_BIT_CU_RBUMP 0b01000000  // CoverUI-RBUMP (as of CoverUI FW 2.1x)
 
 #define LL_STATUS_BIT_UI_AVAIL 0b10000000
```
rovo89: Could you use (1 << n) for these as well? I find the 0b... format very hard to read.

Firmware/LowLevel/src/datatypes.h
```diff
@@ -135,22 +129,67 @@ struct ll_ui_event {
...
+    float v_charge_cutoff = 30.0f;         // Protective max. charging voltage before charging get switched off (-1 = unknown)
+    float i_charge_cutoff = 1.5f;          // Protective max. charging current before charging get switched off (-1 = unknown)
+    float v_battery_cutoff = 29.0f;        // Protective max. battery voltage before charging get switched off (-1 = unknown)
+    float v_battery_empty = 21.7f + 0.3f;  // Empty battery voltage used for % calc of capacity (-1 = unknown)
+    float v_battery_full = 28.7f - 0.3f;   // Full battery voltage used for % calc of capacity (-1 = unknown)
+    uint16_t lift_period = 100;            // Period (ms) for >=2 wheels to be lifted in order to count as emergency (0 = disable, 0xFFFF = unknown). This is to filter uneven ground
+    uint16_t tilt_period = 2500;           // Period (ms) for a single wheel to be lifted in order to count as emergency (0 = disable, 0xFFFF = unknown). This is to filter uneven ground
+    float shutdown_esc_max_pitch = 0.0f;   // Do not shutdown ESCs if absolute pitch angle is greater than this (0 = disable, 0xffff = unknown) (to be implemented)
```
rovo89: Would you use this field both as a general toggle for the shutdown and for configuring the max pitch? My intuition is to better add another boolean field before this. Or uint8 if we need different modes, but that wouldn't impact the struct size.

Firmware/LowLevel/src/main.cpp
```diff
@@ -123,15 +118,36 @@ uint16_t ui_version = 0;                   // Last received UI firmware version
...
 uint16_t ui_interval = 1000;               // UI send msg (LED/State) interval (ms)
 
-// Some vars related to PACKET_ID_LL_HIGH_LEVEL_CONFIG_*
-uint8_t comms_version = 0;  // comms packet version (>0 if implemented)
-uint8_t config_bitmask = 0; // See LL_HIGH_LEVEL_CONFIG_BIT_*
+// LL/HL config
+#define CONFIG_FILENAME "/openmower.cfg"  // Where our llhl_config get saved in LittleFS (flash)
+uint16_t config_crc_in_flash = 0;
+struct ll_high_level_config llhl_config;  // LL/HL configuration (is initialized with YF-C500 defaults)
+static_assert(sizeof(ConfigOptions) == 1, "Enlarging struct ConfigOption to a sizeof > 1 will break packet compatibilty");
```
rovo89: If feel that this should be right under the struct definition, but not sure if including it in the header has disadvantages.

Firmware/LowLevel/src/main.cpp
```diff
@@ -547,13 +561,81 @@ void onUIPacketReceived(const uint8_t *buffer, size_t size) {
...
+                lift_hall_idx++;
+                break;
+            case HallMode::STOP:
+                stop_halls[stop_hall_idx] = i;
+                stop_hall_idx++;
+                break;
+        }
+    }
+    lift_halls[lift_hall_idx] = 0xff;  // End of lift halls marker
+    stop_halls[stop_hall_idx] = 0xff;  // End of stop halls marker
```
rovo89: If you have all 10x hall or 10x stop, this will overflow. I have an idea for the hall handling, but need to write it on my laptop.

Firmware/LowLevel/src/main.cpp
```diff
@@ -140,66 +156,59 @@ void setRaspiPower(bool power) {
...
+    if (stop_pressed) {
+        if (button_emergency_started == 0) button_emergency_started = millis();  // If we just pressed, store the timestamp
     } else {
-        // Not pressed, reset the time
-        button_emergency_started = 0;
+        button_emergency_started = 0;  // Not pressed, reset the time
+    }
+    if (button_emergency_started > 0 && (millis() - button_emergency_started) >= BUTTON_EMERGENCY_MILLIS) {
+        emergency_state |= LL_EMERGENCY_BIT_STOP;
     }
```
rovo89: This should be an "else if" to the "if (button_emergency_started == 0)" above.

Firmware/LowLevel/src/main.cpp
```diff
@@ -140,66 +156,59 @@ void setRaspiPower(bool power) {
...
+    }
 
     // Handle emergency "Stop" buttons
-    if (emergency_read & LL_EMERGENCY_BITS_STOP) {
-        // If we just pressed, store the timestamp
-        if (button_emergency_started == 0) {
-            button_emergency_started = millis();
-        }
+    if (stop_pressed) {
+        if (button_emergency_started == 0) button_emergency_started = millis();  // If we just pressed, store the timestamp
```
rovo89: I personally prefer the comment on its own line, especially if it explains the branch (i.e. the situation which brought us here) more than the particular line. What I mean is that it's obvious that "button_emergency_started = millis()" stores the timestamp, we wouldn't need any comment for that, but the "just pressed" comment *is* helpful because it explains when "button_emergency_started == 0" can happen.

Firmware/LowLevel/src/main.cpp
```diff
@@ -140,66 +156,59 @@ void setRaspiPower(bool power) {
...
     } else {
-        // Not pressed, reset the time
-        button_emergency_started = 0;
+        button_emergency_started = 0;  // Not pressed, reset the time
```
rovo89, line 188: Same here, actually this might be the better example for what I said.

Firmware/LowLevel/src/main.cpp
```diff
@@ -547,13 +561,81 @@ void onUIPacketReceived(const uint8_t *buffer, size_t size) {
...
-    ll_config.config_bitmask = config_bitmask;
-    ll_config.volume = 80;            // FIXME: Adapt once nv_config or improve-sound got merged
-    strncpy(ll_config.language, "en", 2); // FIXME: Adapt once nv_config or improve-sound got merged
-    sendMessage(&ll_config, sizeof(struct ll_high_level_config));
+void sendConfigMessage(const uint8_t pkt_type) {
+    size_t msg_size = sizeof(struct ll_high_level_config) + 3;  // + 1 type + 2 crc
+    uint8_t *msg = (uint8_t *)malloc(msg_size);
+    if (msg == NULL)
+        return;
+    *msg = pkt_type;
```
rovo89: I'd prefer msg[0] to avoid the impression that the whole content is overwritten, what foul you think?

rovo89: Also, I assume that the CRC will be calculated later? Might be worth adding a comment about that.

Firmware/LowLevel/src/main.cpp
```diff
@@ -547,13 +561,81 @@ void onUIPacketReceived(const uint8_t *buffer, size_t size) {
...
+ * @brief applyConfig applies those members who are not undefined/unknown.
+ * This function get called either when receiving a ll_high_level_config packet from HL, or during boot after read from LittleFS
+ * @param buffer
+ * @param size
+ */
+void applyConfig(const uint8_t *buffer, const size_t size) {
+    // This is a flexible length packet where the size may vary when ll_high_level_config struct get enhanced only on one side.
+    // If payload size is larger than our struct size, ensure that we only copy those we know of = our struct size.
+    // If payload size is smaller than our struct size, copy only the payload we got, but ensure that the unsent member(s) have reasonable defaults.
+    size_t payload_size = min(sizeof(ll_high_level_config), size - 2);  // exclude crc
```
rovo89: I think we should subtract 2 where the method is called, so size is really the amount of bytes in the buffer that are relevant for the method.

Firmware/LowLevel/src/main.cpp
```diff
@@ -547,13 +561,81 @@ void onUIPacketReceived(const uint8_t *buffer, size_t size) {
...
+    // It's important to use a clean one with LL-default values, and not the current live llhl_config one,
+    // because we need to ensure that re-commented mower_config parameter get the LL-default value and not the last saved one!
+    ll_high_level_config new_config;
+
+    // Take over of payload values. This could also be made the other way around, which would save the copy of the "leading"
+    // values, but I felt it's more secure to do it this way (in special for further packet enlargements)
+
+    // HL always force these member
+    new_config.options = rcv_config.options;
+    strncpy(new_config.language, rcv_config.language, 2);
```
rovo89: Why do these need to be handled differently? You could one bit to indicate "keep defaults" and the language could be {0,0} for undefined.

Thinking more about it, I'm wondering whether it wouldn't be better to avoid the bitmask completely because it will make extensions harder, not easier. If we add another bit, it would have to be off unset by default, otherwise an older ROS version will always disable it because it will send a 0 in that place. Whereas adding another boolean at the end of the structure should now be no additional effort and you have all freedom to choose defaults.

Or instead of a boolean, there should probably be a tri-value on/off/undefined enum (or 1/0/-1 int8) that can be sent by ROS, following what you did for all the other settings.

Firmware/LowLevel/src/main.cpp
```diff
@@ -583,33 +665,24 @@ void onPacketReceived(const uint8_t *buffer, size_t size) {
...
-            comms_version = pkt->comms_version;
-        else
-            comms_version = LL_HIGH_LEVEL_CONFIG_MAX_COMMS_VERSION;
-        config_bitmask = pkt->config_bitmask; // Take over as sent. HL is leading (for now)
-        // FIXME: Assign volume & language if not already stored in flash-config
+    } else if (buffer[0] == PACKET_ID_LL_HIGH_LEVEL_CONFIG_REQ || buffer[0] == PACKET_ID_LL_HIGH_LEVEL_CONFIG_RSP) {
+        applyConfig(buffer + 1, size - 1);  // Skip type
+
+        // Store in flash
+        saveConfigToFlash();
```
rovo89: This will save the effective config, right? If we later change any defaults, it will still load the old values from flash. I think we need to save the packet from ROS instead. We can even store it in full length and keep the CRC, because applyConfig() will be used when loading it.

Firmware/LowLevel/src/main.cpp
rovo89: How about this: Define `lift_halls` and `stop_halls` as vector, so you don't need the stop bytes and can simply iterate over them with the colon syntax. Instead of storing the indexes, use a struct which has everything you need, i.e. a copy of the HallConfig (actually no, just need `active_low`) and a copy of the function pointer from `halls`.

Something like this:
```c++
struct HallsCache {
  std::function<bool()> is_high;
  bool active_low;
}
std::vector<HallsCache> stop_halls;
std::vector<HallsCache> lift_halls;
```

Use your existing loop to append elements to the vectors and copy over the flag from the config and the function pointer. And then for the iteration:

```c++
bool stop_pressed = false;
for (HallsCache &hall : stop_halls) {
  if (hall.is_high() ^ hall.active_low) {
    stop_press = true;
    break;
  }
}
```

Take the above more as pseudo-code to get the idea, not strictly as implementation suggestion.

Firmware/LowLevel/src/datatypes.h
Apehaenger: Yip. Done.

Firmware/LowLevel/src/datatypes.h
Apehaenger: Yes sure. Bit shifting format is also less error-prone!
Done.

Firmware/LowLevel/src/datatypes.h
Apehaenger: Yes, I really prefer to use one field, because we do it also with some other fields where a value of 0 would be sense-less and thus can effective be used as "off" or "disabled" indicator.

Well, I'm aware of the ugly comparison to zero with floats, and also wondered during takeover of that field, if the "pitch angle" really needs to be a float?! In my opinion a integer should be fully sufficient.

So I would prefer a uint8_t but with 0=disabled?

Firmware/LowLevel/src/main.cpp
Apehaenger: Thought static_assert() don't work in the header :innocent:, also thought that I already tested it in header :innocent: 
But tested again, and it's working! So, "yes", much better place for the static_assert().
Done.

Firmware/LowLevel/src/main.cpp
Apehaenger: Very good hint with the >10 overflow!!

I don't feel good with the std::array nor std::vector due to the heap risk @ embedded.

~~In fact I already tried with etl::array as well as etl::vector but recognized during the implementation that etl::array~~|~~vector.size() and their iterators work over the full initialized size and thus another variable would become required (stop_ & lift_halls_max). So I dropped them again :no_mouth:~~ wrongly remembered!

Firmware/LowLevel/src/main.cpp
Apehaenger: Yes, right. msg[0] is more clear.

Yes CRC is calculated in sendMessage() for all LL packets (heartbeat, UI, status, ...). But sure, better to much comments than to less ;-)
Done

Firmware/LowLevel/src/main.cpp
Apehaenger: Yes, good point. More clear!
Done

Firmware/LowLevel/src/main.cpp
Apehaenger: > Why do these need to be handled differently? You could one bit to indicate "keep defaults" and the language could be {0,0} for undefined.
> 
It's reasoned in some discussion I've had in past during Improve-Sound PR.
That time, it was said that the language should be always given by HL and the same with DFP_IS_5V.
And during switchover from bitmask to bitfields I recognized that's also true for the two new bit- switches.
In addition I couldn't imagine any option, where LL could be leading for an option.

> Thinking more about it, I'm wondering whether it wouldn't be better to avoid the bitmask completely because it will make extensions harder, not easier. If we add another bit, it would have to be off unset by default, otherwise an older ROS version will always disable it because it will send a 0 in that place. Whereas adding another boolean at the end of the structure should now be no additional effort and you have all freedom to choose defaults.
> 
I also thought shortly about a possible new option in the bitfields options. But don't see an issue with the fact that it has to be integrated then, as default=false option :confused: 

Bit masks are pretty effective and and I really do like the more modern and compact implementation via bitfields :heart_eyes: 


> Or instead of a boolean, there should probably be a tri-value on/off/undefined enum (or 1/0/-1 int8) that can be sent by ROS, following what you did for all the other settings.

Hmm... tri-value enums in a larger bitfield :stuck_out_tongue_winking_eye: 
But really?! We don't have any tri-state-option now, nor I can imagine about one, as well as we've now 3 used options out of 8!
Adding additional bools or even tri-state-switches later, when required or to simple-minded (by the implementor) remain a future option :wink:

Firmware/LowLevel/src/datatypes.h
rovo89: I agree that uint8 should be enough. Or int8, if you want -1 for undefined.

For the double use: As long as it's only one setting, I'm fine with it. That other PR isn't finished though, so what if we add something else like a delay?

Firmware/LowLevel/src/main.cpp
rovo89: Is there a way to check whether we're anywhere close to the limits? And are the effects really that bad? Would make the code quite a bit simpler.

I found https://stackoverflow.com/questions/13749096/c-vector-on-arduino-takes-up-massive-ram-footprint, not sure if there would be a way to reserve space for only 10 elements, so it doesn't need to grow. Maybe that would help to keep memory allocation in sensible range?

Firmware/LowLevel/src/main.cpp
rovo89: > That time, it was said that the language should be always given by HL and the same with DFP_IS_5V.

I thought for a long time that everything is leading in HL... 😉 But where's the difference between these two settings and let's say the volume or the max pitch? You have already given a default language, so why shouldn't that be used unless explicitly overridden?

> But don't see an issue with the fact that it has to be integrated then, as default=false option 😕

You can always invert the meaning, sure. But let's say you want do enable background sounds by default. Then the setting would be no_background_sound=false... Double negation.

> We don't have any tri-state-option now, nor I can imagine about one

We will need that when we add boolean options. For the other data types, you can use -1 or 0xff as "undefined" value from the ROS side, but with a boolean, that will be difficult.

Firmware/LowLevel/src/datatypes.h
Apehaenger: Okay, then I would choose uint8_t so that all values get handled similar :wink: 

Delay :scream: ?!
Well I already tested it a couple of weeks ago... so if you expect further additions like delay, inverse, moonshine exception or alpha x-ray pollution, lets better throw it out completely... so it can be decided by whom ever it get integrated when it's ready :innocent: ?

Firmware/LowLevel/src/main.cpp
Apehaenger: > Is there a way to check whether we're anywhere close to the limits? And are the effects really that bad? Would make the code quite a bit simpler.
> 
Not really reliable.
For STM32 it's really somehow risky and guess for Pico it will be similar.
Just had exactly such an issue with my Stock-CoverUI port where I used a dirt tiny button queue from std::.
One guy suddenly had issues with it and after some hours of debugging it turned out that it was the heap overflow triggered by the std::whatever container.

> I found https://stackoverflow.com/questions/13749096/c-vector-on-arduino-takes-up-massive-ram-footprint, not sure if there would be a way to reserve space for only 10 elements, so it doesn't need to grow. Maybe that would help to keep memory allocation in sensible range?

Yes, that's what the [etl](https://www.etlcpp.com/) implementation does. All required memory get reserved during creation and (in my point of view) it's rock solid to use, **but** the size is fixed then and luxury shorthanded iteration is done over the full sized container and not the used one :disappointed: so another var would be needed holding our used-size values and whoops the nice thing becomes ugly.

While writing.... will check my CoverUI code... think I used a etl::queue there and the queue had a usable used-size (independent of the initialized size) if I remember right :thinking:

Firmware/LowLevel/src/main.cpp
Apehaenger: > > That time, it was said that the language should be always given by HL and the same with DFP_IS_5V.
> 
> I thought for a long time that everything is leading in HL... 😉 But where's the difference between these two settings and let's say the volume or the max pitch? You have already given a default language, so why shouldn't that be used unless explicitly overridden?

I know, I used the "leading" phrase the first time related to this topic :innocent: ... it's probably miss leading :thinking: 

Volume could be controlled from both sides, and the user is free to decide if HL is forcing a specific volume or if it's changeable i.e. via CoverUI. Same as with language, it was also configurable from both sides in past, but got changed some months ago.
Both of them needs to be known during boot.
Similar to "max pitch", which should have a reasonable value (if enabled), even if HL is offline. But for "max pitch" we only have HL as controlling side at the moment. Very unlikely, but if one plans to integrate it i.e. into a CoverUI with Display, he probably need to change that.


> 
> > But don't see an issue with the fact that it has to be integrated then, as default=false option 😕
> 
> You can always invert the meaning, sure. But let's say you want do enable background sounds by default. Then the setting would be no_background_sound=false... Double negation.

Yes, that's clear to me, but sound also very theoretical to me. Well background sound would be uncritical because it could wait till it receives a config update from HL, but "yes" if there's a critical option in future, which need to be on by default, the implementer need to use a double negation label.

But imagine:
Isn't "IGNORE CHARGING CURRENT" in real a negated "ENABLE CHARGE CURRENT PROTECTION" ?

> 
> > We don't have any tri-state-option now, nor I can imagine about one
> 
> We will need that when we add boolean options. For the other data types, you can use -1 or 0xff as "undefined" value from the ROS side, but with a boolean, that will be difficult.

Okay, you're thinking about an option LL could have, but HL doesn't has a preference for.
Yip, for such a case we need a tri-state option. 
Will check to change the bitfield from boolean to a tri-state (enum) option.

Firmware/LowLevel/src/main.cpp
Apehaenger: Yes, think you're right! Good hint!!

Apehaenger: Really clever hint!! :+1: 
Committed, not yet tested, but think should work :innocent:

Firmware/LowLevel/src/main.cpp
Apehaenger: Okay, ETL vectors should do it. Not yet tested, but will test again once all is done.

I'm pretty surprised about the costs for this more comfortable container handling.
While adding the structs and the new container stuff, I thought it will cost us at least 1kByte in sum, but at the end, it's only +200byte RAM and +100byte flash. Expected much more :+1: 

Just committed the changes

Firmware/LowLevel/src/main.cpp
Apehaenger: Well the comments are in since about July 2022 :innocent: 
Don't know if Clemens integrated them or someone else.
In any case, I mostly try to not remove any comments from someone else.
Instead of, if I've the impression that the code is pretty clear, I move them to the end of line. By this, the one who struggles with the code, can see a short hint at the end, but it doesn't waste vertical code reading.

Firmware/LowLevel/src/main.cpp
Apehaenger: Done

PR #115, opened 2025-05-01
Drop PioSPI and unpin arduino-pico

olliewalsh:
Pinning ardunio-pico to an older version in 3c5f75103aad2a155e7a7e50a570586d5fd2d655 caused issues with serial I/O on some boards (definitely my 0.11, and similar reports for a 0.13).

Switch back to the current arduino-pico version by replacing PioSPI with SoftwareSPI from arduino-pico.

Tested on 0.12.x board, which uses PIO SPI for the IMU.

### DIFF
--- a/Firmware/LowLevel/platformio.ini
+++ b/Firmware/LowLevel/platformio.ini
@@ -18,12 +18,6 @@ platform = https://github.com/maxgerhardt/platform-raspberrypi.git
 board = pico
 framework = arduino
 
-; Latest (known) version which compiles well with jpiat/PioSPI.
-; framework-arduinopico >= 4.4.3 will have a (PIO-Assembly driven) SoftwareSPI implementation which would make jpiat/PioSPI needless,
-; but I currently have no mainboard with LSM6DSO here for testing. So stick to the last working version till someone can switch & test it.
-platform_packages =
-    framework-arduinopico @ https://github.com/earlephilhower/arduino-pico.git#4.3.0
-
 board_build.core = earlephilhower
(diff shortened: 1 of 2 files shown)


Firmware/LowLevel/platformio.ini
```diff
@@ -18,12 +18,6 @@ platform = https://github.com/maxgerhardt/platform-raspberrypi.git
 board = pico
 framework = arduino
 
-; Latest (known) version which compiles well with jpiat/PioSPI.
-; framework-arduinopico >= 4.4.3 will have a (PIO-Assembly driven) SoftwareSPI implementation which would make jpiat/PioSPI needless,
-; but I currently have no mainboard with LSM6DSO here for testing. So stick to the last working version till someone can switch & test it.
-platform_packages =
-    framework-arduinopico @ https://github.com/earlephilhower/arduino-pico.git#4.3.0
```
olliewalsh, line 25: Maybe we should pin this to the current version? @Apehaenger WDYT?

olliewalsh: Or next year may have the same issue where not much has landed here over the winter so CI didn't run, then we find that builds are broken in the spring

Apehaenger (approved):
Quite cool, that's the better way to go with the new SoftwareSPI implementation :+1: 

I'm fine going with current version instead of pinning to a specific version. We need to recognize incompatibilities in some way and I think as earlier as better. And in this special case only a couple of people had that 0x0 issue.

